PR #48, opened 2019-03-01
Oof

xBelladonna:
Added the missing closing square bracket to `pk;system new [system name` on the help output for `pk;system new`


xBelladonna:
The funny thing is, I did not even do this. Thanks Katie xD

PR #49, opened 2019-03-01
Anti hax

xBelladonna:
Send system json file in DMs instead of publicly


xBelladonna:
F why didn't the changes push

xBelladonna:
Oh god

PR #52, opened 2019-03-05
Added easter eggs

xBelladonna:
Added easter eggs of the bound to Earth variety


xSke:
Could I get you to remove the hint in the help file? It's more fun that way :)

xBelladonna:
[Katie] Just force-pushed the update

PR #53, opened 2019-03-06
Added notice for member avatar image dimensions, fixed timezone command

xBelladonna:
Added notice that member avatars have to have 1 dimension 1024 pixels or less, changed the system timezone command to reflect the new timezone searching by location method.


xBelladonna:
All I did was sync the fork ;~;

PR #54, opened 2019-03-07
Corrects aiohttp error discord-py throws.

Ryonez:
Error when using latest aiohttp:
```Successfully built aiohttp asyncpg discord.py humanize uvloop ciso8601 pycares multidict yarl idna-ssl regex tzlocal websockets numpy cffi pycparser
discord-py 1.0.0a0 has requirement aiohttp<3.5.0,>=3.3.0, but you'll have aiohttp 3.5.4 which is incompatible.
```

Correction is to set version used to 3.5.0.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,5 +1,5 @@
 aiodns
-aiohttp
+aiohttp==3.3.0
 asyncpg
 dateparser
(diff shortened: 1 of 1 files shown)


Ryonez:
Correction is to set version used to 3.3.0, not 3.5.0.

The commit comment says the incorrect version, but the adjustment is correct.

PR #58, opened 2019-03-08
Reduce to 8 members per page and truncate description at 512 characters

xBelladonna:
Literally the title.... need to figure out a newline between the end of the description and the title of the next member but it's 5 AM....


xBelladonna:
Latest commits to upstream discombobulated things.......

PR #64, opened 2019-03-24
Add API endpoint to query current fronters (/s/<hid>/fronters), and more!

xBelladonna:
Querying this endpoint returns the switch timestamp and the member objects of the current fronters :3

### DIFF
--- a/src/api_main.py
+++ b/src/api_main.py
@@ -47,7 +47,7 @@ class Handlers:
     @require_system
     async def get_system(request):
         return web.json_response(request["system"].to_json())
-    
+
     async def get_other_system(request):
         system_id = request.match_info.get("system")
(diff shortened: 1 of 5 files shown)


xBelladonna:
Added more commits because jkdfksjfkdsl make good

xBelladonna:
Made it so that API tokens get sent in DMs on a separate message so it's easier for mobile users to copy and paste

xBelladonna:
Changed the help sections so that command notes get put at the end of pk;help member

xBelladonna:
Made the command_notes get called from misc_commands.py instead of in the member help section of help.py so it doesn't get written twice for no good reason (and also doesn't break the character limit hhhh)

xBelladonna:
Closing to rebase new sharding commit

PR #67, opened 2019-03-28
Update token disclaimer, add checkmark reaction on token command issue

xBelladonna:
Added a warning emoji before token disclaimer in DMs. 

Bot will react with a green checkmark to the pk;token command in the channel (or DM) to let users know it worked.


xBelladonna:
Made change as suggested by @xSke (including a rebase of an upstream commit, that's why the push was forced).

PR #68, opened 2019-03-28
Update api_main.py

greysdawn:
Removing API-side OAuth flow to accept a Discord token from the request


xSke:
Not merging this - it would allow a third-party application owner to spoof any of its users. I'd rather we just do the OAuth auth flow normally.

PR #78, opened 2019-04-18
Add API endpoint to get message

goose121:
This allows, for example, other bots to be aware of the members of a system and to distinguish between them.


goose121:
wait wait wait no this is incorrect stop the presses

PR #91, opened 2019-05-05
bot: add member count to system card

xBelladonna:
Just what the title says

### DIFF
--- a/src/pluralkit/bot/embeds.py
+++ b/src/pluralkit/bot/embeds.py
@@ -79,8 +79,14 @@ async def system_card(conn, client: discord.Client, system: System, is_own_syste
     card = discord.Embed()
     card.colour = discord.Colour.blue()
 
+    all_members = await system.get_members(conn)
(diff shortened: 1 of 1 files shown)


xSke:
Don't have time today to do this myself, but could you change it so it puts the count in parens in the embed title?

xBelladonna:
Sure I'll do it once I'm home from work

xSke:
Bonus: have it display the first 10 or so in short form, then refer to the member list if there are more

xBelladonna:
*sweats nervously*
Sure I'll give it a go

xBelladonna:
Oh god I forgot the commit formatting... I'll just... force push that...

xBelladonna:
Ok, it displays member count in embed title or member field if the system is unnamed now.

xSke:
Could you make it display in the member field title no matter what (and then not in the system name)? Seems more consistent that way.

xBelladonna:
It hath been pushed 👀

PR #95, opened 2019-05-08
removed disk writes

baby-princess:
lowed the number of files

reduced disk writes by removing empty __init__.py files

removed the .vscode dir as it was in the .gitignore

added the example configuration to the readme to tidy the project up a little


xSke:
I really don't understand the motivation for these changes. Could you explain?

baby-princess:
i was originally going to strip the need for docker, because its insecure, and can easily be broken out of the sandbox. and replace it with a bash script though, i never got around to it. As for the less disk writes, having empty files, still requires them to be wrote the disk, which when running on an SSD is fairly limited, also writing empty files is always a bad practice anyways and it makes projects untidy, having a un-needed example config file, could be moved into the README, that's all i read in projects, the alternative would be an INSTALL file, however less files makes projects cleaner. I removed the .vscode dir as it was in the git ignore anyways, also including editor files is never a good idea, as it binds people to that editor, and i always delete the .vscode dirs also, i do not need VScode projects on my system thanks. as for __init__.py's they are mostly empty, and python should generate them its self if need be. moved the shell scripts into one.

Also as a side note, there is no reason to make people need postgress, nor docker, the same result could easily be achieved with a flat database, which is preferred, as it can be used with native core utils, over having to install an extra package or 2, likewise everything ran from docker could be ran from a shell script. The better practice that complies with the UNIX philosophy more would be to remove anything unneeded, and to keep the project simple.

So the main motivation of these changes is to make it more compliment to the UNIX philosophy :).

xSke:
> i was originally going to strip the need for docker, because its insecure, and can easily be broken out of the sandbox. 
Docker isn't supposed to be a secure sandbox, it's for containerizing an application and its ecosystem so it can be deterministically reproduced on different hosts.

> As for the less disk writes, having empty files, still requires them to be wrote the disk, which when running on an SSD is fairly limited
This is negligible in practice, and if you're going for microoptimizations of this degree there are loads more low hanging fruit.

> also writing empty files is always a bad practice anyways and it makes projects untidy
"Bad practice" is a matter of opinion. I'd agree with you that there's no reason to have empty files for no reason, but sometimes there's a point.

> having a un-needed example config file, could be moved into the README
It could be, but then you'd have to open up the README file, copy the specific lines out (not an easy task on CLIs unless you're proficient in vi), and then paste them in, whereas having them in a separate file just makes it a matter of `cp pluralkit.conf.example pluralkit.conf`.

> the alternative would be an INSTALL file, however less files makes projects cleaner.
Matter of opinion, again. Sometimes having more files is cleaner - separation of concerns and all.

> including editor files is never a good idea, as it binds people to that editor
Not really - people can just *not* use them. The included files were just a run configuration, which just makes it easier for people using said editor (probably most potential contributers, including myself) to get started.

> as for init.py's they are mostly empty, and python should generate them its self if need be.
It won't. `__init__.py` files are used for signaling to (C)Python that the directory it's in is considered a module - thus removing the files will break running the program. Not a fan of this design either, but it is how it is.

> moved the shell scripts into one.
Having them separate makes them easier to call and to get an overview of what they are and do.

> Also as a side note, there is no reason to make people need [Docker]
There are instructions on running the bot without Docker in the README file.

> [there is no reason to make people need] postgres, the same result could easily be achieved with a flat database, which is preferred
Databases exist for a reason. Implementing the code necessary to duplicate the functionality using the database offers would be a monumental task, and just using a pile of .json files in lieu of a real database for an application that actually benefits from the perks a database offers is an awful idea for numerous reasons.

Since your pull request offers very little in terms of actual useful changes, I won't be merging this pull request.

baby-princess:
no you would not need a text editor at all, again coreutils can help, sed awk, grep and echo would allow you to grab the information you need, if you think people actully open up readme's in a text editor, they don't for the most part, the only time i ever open a readme in a editor, is if i'm actually editing it. generally i don't even look through them what so ever, because the man page helps me if need be.

I'm aware that Docker is a containerize environment, however it is supposed to run in a sandbox in order to containerize said solutions.

flat file databases can be managed using coreutils, and it would still be one file, also it should be noted that indexing through awk and sed is far faster then SQL queries (i know this because I've used both for years, and at work we stopped using large databases for a few shell scripts, because its faster and easier to manage). 

I am aware about the alternative instructions in the README.

hm then in that case the init.py system is pretty stupid.

Agreed yes it is my opinion, however on every git repo i download, i always check for a Makefile, setup.py, INSTALL or a CmakeLists.txt, because then i don't have to even touch the README.

As a side note for comparability with other distros, inorder to make it distro independent, like you wish, currently you are bound to anything but Gentoo or Slackware, pip should be ran with --user option so `pip install --user -r requirements.txt` this will be fine to run on any distro.

PR #97, opened 2019-05-14
Added more robust avatar URL processing

direshawarma:
Avatar URLs are now:
1. checked for HTTP/HTTPS schema (already present feature)
2. checked for non-200 response codes
3. checked for acceptable content type
4. checked for size < 1000000 bytes

### DIFF
--- a/src/pluralkit/utils.py
+++ b/src/pluralkit/utils.py
@@ -6,9 +6,13 @@ import string
 from datetime import datetime, timezone, timedelta
 from typing import List, Tuple, Union
 from urllib.parse import urlparse
+from urllib.request import urlopen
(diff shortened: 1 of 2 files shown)


src/pluralkit/utils.py
```diff
@@ -6,9 +6,13 @@
 from datetime import datetime, timezone, timedelta
 from typing import List, Tuple, Union
 from urllib.parse import urlparse
+from urllib.request import urlopen
+from urllib.error import HTTPError
+import requests
```
xSke: no.

xSke:
The `master` branch is currently in maintenance mode, and the equivalent changes have already been implemented in the rewrite branch.

PR #100, opened 2019-06-06
Add proxy tags to full system list

xBelladonna:
Adds a field titled `Proxy tags:` to the embed with prefixes and suffixes displayed as: [`text`] or Jane:`text`


xSke:
Can you put the brackets inside the code block?

xSke:
Can you put the brackets inside the code block?

xBelladonna:
- Put proxy tags inside code block so it display as such: `[text]` or `Jane: text`

PR #102, opened 2019-06-09
Add fallback to export when user has DMs disabled

xBelladonna:
- Try to DM the user with the export file, and fall back to posting it in the channel if forbidden.
- Uses reactions to confirm whether the user wants to have it posted in the channel and cancels the export if they decline


xBelladonna:
- Allow the user to delete the exported data from the channel using reactions

PR #104, opened 2019-06-15
Fix @mention code injection bug

xBelladonna:
- The bot throws an error on unknown command with the first argument echoed in the reply. This fixes code injection of mentions into the string by regex matching and inserting a zero-width space.
- Ideally I'd convert bot replies into embed cards but this is more of a patch until the rewrite is finished.


xBelladonna:
Son of a.... force pushed and forgot I had this PR open 😂

PR #105, opened 2019-06-15
Delete bot posts by reaction

xBelladonna:
- Enables users to delete the posts the bot makes by reaction
  - When PluralKit posts a large embed like a system card, member card or member list (essentially any embed), they will also react with 🗑 and if the user clicks it, the embed and the original message with the command in it is deleted from the channel. 
  - If the bot doesn't have the `Manage Messages` permission, then it silently fails, the user can then delete their own message if they want

### DIFF
--- a/src/pluralkit/bot/commands/__init__.py
+++ b/src/pluralkit/bot/commands/__init__.py
@@ -164,6 +164,16 @@ class CommandContext:
         except asyncio.TimeoutError:
             raise CommandError("Timed out - try again.")
 
+    async def delete_by_react(self, user: Union[discord.Member, discord.User], message: discord.Message):
+        try:
+            await message.add_reaction("🗑")  # Wastebasket
+            reaction, _ = await self.client.wait_for("reaction_add", check=lambda r, u: u.id == user.id and r.emoji in ["🗑"], 
+                                                     timeout=60 * 10) # 10 minute timeout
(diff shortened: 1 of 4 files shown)


xSke:
Awaiting a reaction like that times out rather quickly - and if we removed that, I can see it quickly causing a lot of hanging tasks to spin up. I'd rather we work out something else, like an on-reaction-add handler similar to how it handles the ❌ / ❓ reactions to proxied messages. You'd need a map of bot responses to sender/original somewhere, but that's a lot more lightweight than a complete coroutine.

xBelladonna:
H yeah that's why I set the timeout to 10 minutes, figured if they didn't delete it before then, they never will 😂 
I can whip the thing up like you said and force-yeet it if you want?

xSke:
Honestly, I think it's better to just leave it to the rewrite branch, if you ask me. But if you implement it and test it i'll `yeet` :)

xBelladonna:
Yes my main concern is getting the python branch """complete""" before we zucc it for the rewrite grhjghdfj but thanks I'll do it when I'm not lazy

xBelladonna:
Yeah this is going to require modifying the DB (well I *could* do it another way but it would be rly nasty) and you're right, I'd rather just leave this as it is and do it in the rewrite.

PR #106, opened 2019-06-26
Fix errors and stuff

xBelladonna:
Fixed a bug probably and added errors for pk;log not that we need them but it was suggested so here it is fdhjghdksj


xBelladonna:
I'm gonna close this because rewrite Soon™

PR #110, opened 2019-07-02
Fix missing RedQuestion emoji

BakaKiller:
Help used RedQuestion emoji but it had not been initialized in Emojis class


xSke:
Could you replace it with the escape code? Should be `\u2753`. Just for consistency's sake :)

BakaKiller:
I amended ! I wasn't sure how to find its escape code uwu

PR #115, opened 2019-07-27
Add endpoint to create new member by POST request

xBelladonna:
- Add the endpoint as above
- Handle collisions while generating hid for new systems and members
- Fix a typo that caused a sequence in the markdown docs to be parsed as a tag instead of escaped


xSke:
The HID generation collision solution is susceptible to race conditions. It probably doesn't matter much in the real world, since the worst-case outcome is it throws an exception (as it already does).

PR #118, opened 2019-08-04
Escape markdown in proxy tags on member card

xBelladonna:
- Escape markdown formatting characters in proxy tags when displaying them on the member card
- Fix a typo in the docs


xBelladonna:
I'm not evil I swear

xSke:
...I'm not merging that

xBelladonna:
How about how?

PR #123, opened 2019-08-26
Fix importing brackets from Tupperbox

xBelladonna:
Tupperbox does not set null brackets to null, instead sets to an empty string
  - If brackets are empty string instead of null/undefined (thanks JS for having 2 keywords for the same thing /s), set to null


xBelladonna:
Forgot to rebase my own dang bugfix from the last PR ~~which hasn't been merged either gfdhgfdgh~~

xBelladonna:
Or... that wasn't actually necessary to do anyway... fshgfdhdgf I'm tired just @ me if it's wrong and I need to fix it
image

xSke:
Messy merge (and I'm not 100% sure the fix would work just looking at it) so I implemented it myself.

PR #128, opened 2019-09-29
Add displayname to PK export/import

nephanim:
- Add display_name property to member on PK export
- Set display_name if provided when importing from PK export file


nephanim:
I have import for switches as well, just going to close this one and open another with displayname + switches since this portion is so small.

PR #134, opened 2019-10-20
Add members per system cap & improve import performance

nephanim:
A given system can now have up to 1000 members. Within 50 members of that limit, a warning will display whenever a new member is created via the bot. Once the limit is reached, a final warning will appear indicating that no additional members can be created unless members are first deleted. Attempting to create a new member at that point by any method will result in an error message indicating that the limit has been reached. The API will reject new members once this limit is reached, but doesn't have any sort of early warning threshold like the bot does.

Respecting this in pk;import required some restructuring to tease apart which members already exist and which ones need to be created prior to creating any members as it seems preferable to fail early and give the user the ability to intervene rather than pushing the system to the member cap and requiring manual deletion of "lower priority" members before others can be created. One consequence of the restructure is that existing members are now being read in bulk which is a performance improvement of 25-70% depending on how many members/switches need to be imported (the more members you have, the more noticeable this is).

This also fixes a command rewrite bug where the import command response messages ("new system" vs "added/changed") are swapped since I stumbled across it in testing.


nephanim:
I've added bulk import for switches to this for the import command.

### **Performance results**

Tests were run on a database with the following characteristics:
- 740k total systems
- 300k total members
- 1.8mm total switches
- 4.3mm total switch_members records

I used a system with the following characteristics:
- 150 members
- 18k switches
- 900k switch_members records

**Before:**
pk;import (as of current master) - 125.92s

**After:**
pk;export - 34.05s (~72% improvement)

I also tested with smaller data sets including my own system's with similar results.

PR #137, opened 2019-10-28
Fix import from Tupperbox regression defect

nephanim:
When converting a Tupperbox import file to the PluralKit format on import, this sets the Tupper's name as the DataFileMember ID which seems like a reasonable default. I'm also initializing the switches collection (even though it's going to be empty in this case).

This fixes a regression defect introduced when switch import was implemented.


nephanim:
Updated to make the DataFileMember ID a GUID. Technically Tupperbox doesn't allow members with the same name, but just because it _shouldn't_ happen doesn't mean it can't... may as well guarantee uniqueness.

PR #138, opened 2019-10-29
Move hid generation into database functions

nephanim:
The hid columns on systems and members tables are now generated automatically on insert. Much like their primary keys, the application can now trust the database to provide these values.

* Adds three new functions to the database:
   - generate_hid - equivalent of the GenerateHid utility method we had before, makes a new hid
   - new_system_hid - calls generate_hid in a loop until unique across systems
   - new_member_hid - calls generate_hid in a loop until unique across members
* Sets the default for hid columns on members and systems tables to use the functions above
* Removes while loops from create system/member methods in favor of letting the database handle the IDs
* Removes the now unused GenerateHid utility method

### DIFF
--- a/PluralKit.Core/Stores.cs
+++ b/PluralKit.Core/Stores.cs
@@ -346,15 +346,9 @@ namespace PluralKit {
         }
 
         public async Task<PKSystem> CreateSystem(string systemName = null) {
-            string hid;
-            do
-            {
-                hid = Utils.GenerateHid();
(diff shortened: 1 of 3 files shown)


nephanim:
Given the low priority of this change, it should be reworked to build out a real migration framework for better separation of concerns between db_schema (which should represent the current state of the database) and a series of update scripts (numbered and run in sequence based on the current version of the database, which is tracked at a global level). Merging this as is would trade one type of technical debt for another. To be revisited soon!

Issue #111, opened 2019-07-14
Add pagination to front history card


xSke:

xSke:
Closed by https://github.com/xSke/PluralKit/commit/b347d2d5576570f349b8b7c7e5a1230d8339bcaa.

PR #139, opened 2019-10-29
Add pagination to front history card

nephanim:
When viewing front history for a system, results are now displayed in pages of 10 instead of displaying only the most recent 10. This addresses #111.

The implementation now uses the same logic as the frontpercent command for determining switch end times to reduce duplicated code and the number of database round trips.

### DIFF
--- a/PluralKit.Bot/Commands/SystemCommands.cs
+++ b/PluralKit.Bot/Commands/SystemCommands.cs
@@ -176,6 +176,16 @@ namespace PluralKit.Bot.Commands
         {
             if (system == null) throw Errors.NoSystemError;
 
+            // Determine page size
+            string pageSizeStr = ctx.RemainderOrNull();
+            int.TryParse(pageSizeStr, out int itemsPerPage);
+            // null or negative page size inputs default to 10 items per page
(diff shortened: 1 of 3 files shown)


nephanim:
Updated front history pagination to incorporate the fix for the embed exceeding the max description length and made page size an optional parameter in so doing. If unspecified, the command defaults to 10 items per page.

xSke:
Hm, i feel kind of weird about just arbitrarily capping each line's length individually. I was playing around with having something that'd dynamically just show as many rows as it can in each page (capped at, say, 20), but that would either need it

nephanim:
I went down that path a bit too, but didn't get far enough to like where it was going. I'll think about it some more when I have some time and see what we come up with. At a minimum I think this could build out the page at full length and do something smarter where it starts chopping portions of lines until they're small enough. I do think it would be better for page size to be based on how much content is shown in a predictable way rather than items if we can arrive at something generic enough that it can be reused easily.

xSke:
Ended up writing this myself after forgetting about this PR, sorry!
https://github.com/xSke/PluralKit/commit/b347d2d5576570f349b8b7c7e5a1230d8339bcaa

PR #140, opened 2019-10-30
Send error if both prefix and suffix are empty when changing proxy tags

berries-boop:
If someone tries to change their proxy tags to `text`, it will clear them (but send a message saying it "changed them to `text`". This can be confusing for users who may not understand how to set proxy tags.
Instead, send an error with a link to the proxying guide, with a sidenote on how to clear the proxy tags if that is the requested action.


xSke:
Solved (ish) by https://github.com/xSke/PluralKit/commit/5caf040bbcd39e7ebd5f5b6b2608c973d71934f0

PR #141, opened 2019-10-30
Add support for multiple proxy tags

### LINKED ISSUE #103: Allow proxies to be case insensitive
My friend has an issue where, on their phone, it keeps capitalizing the first letter automatically, so,
`a,text` becomes `A,text` while on the phone.

We've worked around this issue by duplicating each member and having one proxy with uppercase and one with lowercase, but it would be nice with a builtin feature to enable case insensitive proxies.

One way would be to enable regex proxies, which would also enable more complex proxies as a whole. For instance we use `text -fs` proxies, and with regex support we could use both `text -fs` and `text - fs`. But that's not as important to us specifically, just voicing this as an option.

There are security/performance concerns with [user specified regex though](https://www.regular-expressions.info/redos.html), but like that site suggests, [we could use a non-backtracking regex engine](https://github.com/google/re2/wiki/WhyRE2) to mitigate the security concerns.

If we don't go with the regex route, I'm thinking a system-wide setting for case-sensitivity, with member specific overrides, in case one member doesn't like it or something, or is the only one who wants it or something.

I'm willing to implement this myself and make a PR but since there's a discussion to be had about how to implement it, I'm creating an issue first.

xSke:
Tangentially closes #103.

Required database migration:
```sql
create type proxy_tag as (
    prefix text,
    suffix text
);

alter table members add column proxy_tags proxy_tag[] not null default array[]::proxy_tag[];
update members set proxy_tags = array[(prefix, suffix)]::proxy_tag[] where prefix is not NULL or suffix is not NULL;
alter table members drop column prefix;
alter table members drop column suffix;
```

### DIFF
--- a/PluralKit.API/Controllers/MemberController.cs
+++ b/PluralKit.API/Controllers/MemberController.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using PluralKit.Core;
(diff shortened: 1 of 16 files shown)


xSke:
Ideally, yes, but I don't have the infrastructure to track the current "version of the database" and automatically apply upgrades as necessary. I only include "creation" commands in the `db_schema.sql` file, and do the rest manually.

PR #142, opened 2019-10-30
Add migrations for proxy_tags and keep_proxy changes

nephanim:
Adds missing database migrations for the following updates:
- members.proxy_tags added
- update members.proxy_tags with combination of members.prefix and members.suffix
- members.prefix removed
- members.suffix removed
- members.keep_proxy added

NOTE: Longer-term, this type of thing should move into a database migration framework instead, but for now these updates should be idempotent and safe to run on each startup. If the migration has already been run, these just won't run / won't do anything.

This is higher priority than #138 which I'll need to update if this is merged in.

### DIFF
--- a/PluralKit.Core/db_schema.sql
+++ b/PluralKit.Core/db_schema.sql
@@ -86,3 +86,34 @@ create table if not exists servers
     id          bigint primary key,
     log_channel bigint
 );
+
+/* UPDATES */
+
+--Add proxy_tags column to members
+ALTER TABLE members ADD COLUMN IF NOT EXISTS proxy_tags proxy_tag[] NOT NULL DEFAULT array[]::proxy_tag[];
+
+--Migrate legacy proxy tags to proxy_tags array
+DO $$ BEGIN
+	IF EXISTS (
+		SELECT column_name 
+		FROM information_schema.columns 
+		WHERE table_schema = 'public' AND table_name = 'members' and column_name = 'prefix'
+	)
+	AND EXISTS (
+		SELECT column_name 
+		FROM information_schema.columns 
+		WHERE table_schema = 'public' AND table_name = 'members' and column_name = 'suffix'
+	)
+	THEN
+		UPDATE members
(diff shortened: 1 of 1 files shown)


xSke:
I'm a bit conflicted on whether or not to merge this and keep the trend going. I'd like to keep `db_schema.sql` focused on the *current* schema, and not essentially include the entire schema's history in increasingly harder-to-read piles of idempotent SQL statements. The proposed solution in this PR seems unclean, somehow, although I can't quite articulate any coherent reasons to deny it.

I know there are third-party instances running, which is why I've started noting necessary database migrations in extended commit descriptions, although I recognise they're difficult to notice there. A different stopgap solution might be having a `MIGRATIONS.md` describing the database schema migrations necessary, including date information and such, but still require them to be run manually.

As for long-term solutions, thoughts on having a "global data" table with a single row, containing the current database version (among other global data)? The current value would be read at startup and numbered migration steps would be applied to bring the version to the latest one (in practice, probably just a folder of `4.sql`, `5.sql`, `6.sql`, etc).

nephanim:
The proposed long-term solution is in line with what my expectation would be, where on startup you determine the database version and run the remaining updates to get to the latest version. I think we're on the same page there and we agree having a massive SQL file that continues to grow as the database changes isn't a permanent solution. If this _were_ to be merged in, I would expect an issue to be logged to get that technical debt addressed, at which point anything that is not an essential part of the current schema can be ripped out of here and into that framework. I'd be happy to take that. I'm also fine if we draw that line in the sand now and close the PR, it just means potential inconsistencies and headaches in the meantime.

My biggest concerns with encouraging everyone to run the migrations manually are:
- It's easy to miss one (less so if they're pulled into documentation and that's maintained) which isn't noticed until someone runs a command that depends on the missed change
- It can introduce drift where some instances of PluralKit are just slightly more of a snowflake than the application is aware of because someone did something differently

nephanim:
To be addressed as part of a longer term framework improvement. Closing for now.

PR #143, opened 2019-11-09
Fix POSTing members with null proxy tags

greysdawn:
Changes the prefix/suffix checking to allow null prefix and/or suffixes in member POST requests


greysdawn:
Fixed by https://github.com/xSke/PluralKit/commit/0f2228582404ea80df161168d4d7dda9406b3716 (I think)

PR #146, opened 2020-01-08
Add pk;random command

berries-boop:
(Pull request re-created, was not able to merge commits)

This command chooses a random member from the sender's system, and shows the member card~

### DIFF
--- a/PluralKit.Bot/Commands/MemberCommands.cs
+++ b/PluralKit.Bot/Commands/MemberCommands.cs
@@ -68,7 +68,7 @@ namespace PluralKit.Bot.Commands
             var members = (await _data.GetSystemMembers(ctx.System)).toList();
             if (members == null || !members.Any())
                 throw Errors.NoMembersError;
-            var randInt = randGen.Next(members.Count)
+            var randInt = randGen.Next(members.Count);
             await ctx.Reply(embed: await _embeds.CreateMemberEmbed(ctx.System, members[randInt], ctx.Guild));
 
         }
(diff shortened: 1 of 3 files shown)


PluralKit.Bot/Commands/MemberCommands.cs
```diff
@@ -58,6 +58,21 @@ public MemberCommands(IDataStore data, EmbedService embeds, ProxyCacheService pr
...
         }
         
+        public async Task MemberRandom(Context ctx)
+        {
+            ctx.CheckSystem();
+
+            var randGen = new System.Random(); 
+            //Maybe move this somewhere else in the file structure since it doesn't need to get created at every command
+
+            var members = await _data.GetSystemMembers(ctx.System);
```
xSke: GetSystemMembers returns a `IEnumerable`, so you need to wrap it in parens and then call `.ToList()`.

PluralKit.Bot/Commands/MemberCommands.cs
```diff
@@ -58,6 +58,21 @@ public MemberCommands(IDataStore data, EmbedService embeds, ProxyCacheService pr
...
+            if (members == null || !members.Any())
+                throw Errors.NoMembersError;
+            var randMember = members.OrderBy(x=> randGen.Next()).First();
```
xSke: Since `members` is now a list (after calling `.ToList()`), you can just get a random index and pick that one instead of sorting the entire enumerable.

berries-boop:
I really shouldn't be using GitHub as a build testing platform, sorry... 

but it's working finally :D

PR #151, opened 2020-02-07
Improve Docs

BeeFox-sys:
I don't know why that first commit is there and I can't get rid of it, but it doesn't seem to actually change anything so it can really be ignored. The date also seems to me messed up due to the first commit but I have no way of fixing it

Added the getting started page
Changed some urls in the bot to this new page


xSke:
Have you tested this locally? I know the Ruby environment is a bit hard to wrangle but it looks like there are some things that won't render correctly.

BeeFox-sys:
Yep! Testing as I go, the page (Though the rest of the docs have issues due to domain stuff) is up on [git.beefox.xyz/PluralKit/start]http://git.beefox.xyz/PluralKit/start)

BeeFox-sys:
Or apparently it does not like my local testing. Maybe it is domain issues again

PR #154, opened 2020-02-18
Implement permissions checking

0xlgs:
**TL;DR:** Give users more information on what permissions they are lacking when executing a command.

**Important Changes:**
- Every `ctx.Reply` call that sends an embed must be prefixed by `if (await MiscUtils.EnsureProxyPermissions(ctx, "a short description of the action")`. 
- `ctx.PromptYesNo` now takes an input of a string instead of an `IUserMessage`, then either sends the message or edits it to ask the user to respond via message. (calls `ConfirmWithReply`)
- I don't like special cases in shared functions, so instead of having a completely separate description message, `LinkSystem` throws an error on insufficient permissions instead of being moved to text-based confirmation.

**A couple other things:**
- This should also be on anything that needs to be paginated. Maybe add a check in `ctx.Paginate` instead of before every call?
- `ProxyService.EnsureBotPermissions` was moved to `MiscUtils` and renamed `EnsureProxyPermissions`. (And also cleaned up a little.)


0xlgs:
Now that I think of it, there's definitely a few embeds I've forgotten to add the check to. I'll add those in later~

0xlgs:
Closing as this is (very likely) outdated.

PR #158, opened 2020-03-08
Add command to refresh system cache

0xlgs:
Useful when updating information through the API.


xSke:
Was accidentally closed when renaming branches, and I can't reopen due to the original branch being deleted. However, this PR is obsolete as we no longer cache information in the bot itself (due to optimizations in the database later) :)

PR #169, opened 2020-05-10
Add workaround for DSharpPlus/DSharpPlus#575

ParanormalVibe:
This pull request works around an issue with caching within DSharpPlus that rendered some Pluralkit users unable to proxy messages in certain servers. It's possible for the state of the member cache to fall out of sync with the state of the guild cache, causing a null reference exception when `DiscordChannel.PermissionsFor(member)` is called on an affected member, so long as the affected member isn't the server's owner.

I haven't found a way to reliably reproduce this, but a workaround can be achieved by using reflection to validate a member's internal cache before calling `DiscordChannel.PermissionsFor(member)`, which is what I've done here.

### DIFF
--- a/PluralKit.Bot/Utils/DiscordUtils.cs
+++ b/PluralKit.Bot/Utils/DiscordUtils.cs
@@ -1,4 +1,7 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


ParanormalVibe:
In reference to DSharpPlus/DSharpPlus#575

xSke:
Thank you so much for this omg

PR #172, opened 2020-05-11
Feature/proxy command

BeeFox-sys:
This is probably a little messy but I think it works well.

Changelog:
- Removed `pk;proxy` as alias for `pk;system proxy`
- Added `pk;proxy <member> <message>` to allow users to proxy any member in their system, whether they have proxy tags or not


spiralw:
This now conflicts with the `pk;proxy <enable|disable>` command. I still like this feature, not sure where it could go though~

spiralw:
This PR is somewhat outdated, and I'm not quite sure replacing an existing command is the correct thing to do here, but I'd still like to see this - #381 is open for discussion on how that might best be implemented.

PR #173, opened 2020-05-11
qol/private display name

BeeFox-sys:
A small pr that shows the display name of a member in place of their name, where the member is private and the user looking up the member does not have permission to view the full profile


BeeFox-sys:
if #174 is merged, this pr will be irrelevant and can be closed

BeeFox-sys:
Closing; implemented in granular member privacy

PR #174, opened 2020-05-12
Feature/granular member privacy

BeeFox-sys:
This pr will make member privacy granular, as well as add a qol feature to system privacy, being able to set all fields at once with pk;s privacy all <public/private>

There are now 9 settings that can be changed for pk;member <member> privacy.
- name
- description
- birthday
- pronouns
- colour
- date created
- message count
- visibility

Most of these are self explanatory, setting them to private will hide the field, setting them to public will show the field. However two may need some more explaining.
Setting name to private will use the display name of a member, but only if they have one. In which case it will default back to their name.
Setting visibility to private will remove that member from the member list, and from the pk;random command

There are two things that should be improved. ~~That being that member list full currently shows private information for members who are visible, but other things are set to private, when an account linked to the system calls it (other people calling up the full system list will not see private fields. Ideally private fields will not be shown unless the all flag is used, as that also shows hidden members)~~ This has been fixed. And that the user guide needs to be rewritten.

The api has yet to be tested ~~deities please help me my computer will just not let me connect to the api~~ however it runs, and minor changes being made to it should mean there is little issue. It is still recommended that the api be tested before deploying this branch.


acw0:
Did some local testing re: conversation in #api-support, and found the following issues:

- creation date privacy is not validated on the api
- message count privacy does not appear as a setting in the api
- none of the privacy settings seem to be able to be edited through the api

(Also some of the informational messages in the bot are somewhat unclear)

I'll try to get the API features working later this week if IRL stuff lets me~

BeeFox-sys:
Those issues are hopefully fixed, deities I wish I could test this on my computer

BeeFox-sys:
Api has now been tested and all seems to be in order. Only thing left is to write docs

BeeFox-sys:
Documentation is now completed! Everything is tested! Ready for merge

PluralKit.Bot/Commands/MemberEdit.cs
```diff
@@ -362,29 +360,127 @@ public async Task Privacy(Context ctx, PKMember target, PrivacyLevel? newValueFr
...
-            if (newValue == PrivacyLevel.Private)
-                await ctx.Reply($"{Emojis.Success} Member privacy set to **private**. This member will no longer show up in member lists and will return limited information when queried by other accounts.");
+
+            await _data.SaveMember(target);
+            //Handle "all" subject
+            if(subjectStr == "all"){
+                if(levelStr == "private")
+                    await ctx.Reply($"All {target.Name}'s privacy settings have been set to **{levelStr}**. Other accounts will now see nothing on the member card");
+                else 
+                    await ctx.Reply($"All {target.Name}'s privacy settings have been set to **{levelStr}**. Other accounts will now see everything on the member card");
```
xSke, line 475: Nitpick: need periods at the end of these messages, as well as sanitization on target.Name.

PluralKit.Bot/Commands/MemberEdit.cs
```diff
@@ -362,29 +360,127 @@ public async Task Privacy(Context ctx, PKMember target, PrivacyLevel? newValueFr
...
+            } 
+            //Handle other subjects
             else
-                await ctx.Reply($"{Emojis.Success} Member privacy set to **public**. This member will now show up in member lists and will return all information when queried by other accounts.");
+                await ctx.Reply($"{target.Name}'s {subjectStr} has been set to **{levelStr}**. Other accounts will now {levelExplanation}");
```
xSke, line 479: Need period and sanitization here too.

PluralKit.Bot/Commands/SystemList.cs
```diff
@@ -21,7 +21,7 @@ public SystemList(IDataStore data)
         }
 
         private async Task RenderMemberList(Context ctx, PKSystem system, bool canShowPrivate, int membersPerPage, string embedTitle, Func<PKMember, bool> filter,
-                                            Func<DiscordEmbedBuilder, IEnumerable<PKMember>, Task>
+                                            Func<DiscordEmbedBuilder, IEnumerable<PKMember>, Boolean, Task>
```
xSke, line 24: `Boolean` -> `bool`

PluralKit.Core/Migrations/6.sql
```diff
@@ -0,0 +1,23 @@
+-- Create new columns --
```
xSke, line 1: Would like date at the migration header (although I can do this once merged)

docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
     restart: always
   api:
     image: pluralkit
+    # build: .
```
xSke, line 21: Needs to be removed too, unnecessary diff

docs/2-user-guide.md
```diff
@@ -449,15 +449,45 @@ For example:
 When the **member list** is **private**, other users will not be able to view the full member list of your system, but they can still query individual members given their 5-letter ID. If **current fronter** is private, but **front history** isn't, someone can still see the current fronter by looking at the history (this combination doesn't make much sense).
 
 ### Member privacy
-There is also an option to mark a specific member as private, using the command:
+<!-- There is also an option to mark a specific member as private, using the command:
```
xSke, line 452: Are you sure comments like these work? They might still get rendered by Jekyll (if you tested this, nevermind)

docs/2-user-guide.md
```diff
@@ -449,15 +449,45 @@ For example:
...
 
     pk;member <name> private
 
 A private member will *not* be displayed in member lists (even if the member list is public), and will show only limited information if looked up by others - namely name, display name and avatar. Other information, such as description, pronouns and birthday will be hidden.
 
 All of this can only be accessed using the member's 5-letter ID, which is exposed when proxying. So, if you want to keep a member absolutely private, it's recommended you don't proxy with it publicly - that way the ID isn't exposed.
 
-An example of a private member is `cmpuv` - try looking it up and see what's shown, as well as the corresponding system list (`pk;system exmpl list`).
+An example of a private member is `cmpuv` - try looking it up and see what's shown, as well as the corresponding system list (`pk;system exmpl list`). -->
+There is also 9 options for configuring member privacy.
```
xSke, line 461: are also*
nine*
colon instead of period

docs/2-user-guide.md
```diff
@@ -449,15 +449,45 @@ For example:
...
+
+    member <member> privacy <subject> <level>
+
+where `<member>` is the name or the id of a member in your system, `<subject>` is either `name`, `description`, `birthday`, `pronouns`, `color`, `created`, `messages`, or `visiblity` corresponding to the options above, and `<level>` is either `public` or `private`. `<subject>` can also be `all` in order to change all subjects at once.
+
+For example:
+
+    pk;member John privacy visibility private
+    pk;member "Craig Johnson" privacy description public
+    pk;member Robert privacy privacy color public
```
xSke, line 488: privacy privacy?

docs/4-api-documentation.md
```diff
@@ -59,12 +59,20 @@ The following three models (usually represented in JSON format) represent the va
...
 |birthday|date?|Yes|ISO-8601 (`YYYY-MM-DD`) format, year of `0001` or `0004` means hidden year. Birthdays set after 2020-02-10 use `0004` as a sentinel year, but both options are recognized as valid.|
-|prefix|string?|Yes|Deprecated. Use `proxy_tags` instead.|
-|suffix|string?|Yes|Deprecated. Use `proxy_tags` instead.|
+|prefix|string?|Yes|**Deprecated.** Use `proxy_tags` instead.|
+|suffix|string?|Yes|**Deprecated.** Use `proxy_tags` instead.|
 |proxy_tags|ProxyTag[]|Yes (entire array)|An array of ProxyTag (see below) objects, each representing a single prefix/suffix pair.|
 |keep_proxy|bool|Yes|Whether to display a member's proxy tags in the proxied message.|
-|created|datetime|No||
-|privacy|string?|Yes|Patching with `private` will set it to private; `public` or `null` will set it to public.|
+|created|datetime|No|
```
xSke, line 66: IIRC it needs the extra | (blank cell) to handle the table formatting, I may be wrong

docs/4-api-documentation.md
```diff
@@ -59,12 +59,20 @@ The following three models (usually represented in JSON format) represent the va
 |color|color?|Yes|6-char hex (eg. `ff7000`), sans `#`.|
 |avatar_url|url?|Yes|Not validated server-side.|
 |birthday|date?|Yes|ISO-8601 (`YYYY-MM-DD`) format, year of `0001` or `0004` means hidden year. Birthdays set after 2020-02-10 use `0004` as a sentinel year, but both options are recognized as valid.|
-|prefix|string?|Yes|Deprecated. Use `proxy_tags` instead.|
-|suffix|string?|Yes|Deprecated. Use `proxy_tags` instead.|
+|prefix|string?|Yes|**Deprecated.** Use `proxy_tags` instead.|
+|suffix|string?|Yes|**Deprecated.** Use `proxy_tags` instead.|
```
xSke, line 63: Nice, but should be a separate commit/PR (or I can just... do it one off)

docs/2-user-guide.md
BeeFox-sys: comments like that do get rendered as in they are put into the Dom, but they are put into the DOM as is. which means they are still invisible to the end user as the comment is an html comment. however that should have been removed anyway and I'll go remove it when fixing all the reviews

docs/4-api-documentation.md
BeeFox-sys: I'll have another look at the docs but I don't remember running into any issues with it

BeeFox-sys:
thanks for the reviews! I'll hop right on them, and nitpickieness is how you keep a consistent (ish) codebase so dw about it, shouldn't take more then an hour

xSke:
As for actual feature comments, I'm not sure color privacy is really needed. Also unsure about creation date privacy. Would make things simpler to remove those two, thoughts? (not opposed to keeping them either but fewer options = simpler).

Also, I think either name privacy should be removed entirely, or a message should explicitly notify them that name privacy only applies when a display name is set, *and* they'll still need to use the set name for commands. Otherwise people are going to get confused or misled...

docs/4-api-documentation.md
BeeFox-sys: Does not need it c:

docs/4-api-documentation.md
BeeFox-sys: I think one off would be better as it is a change in order to make it constant with the deprecation of the privacy field

BeeFox-sys:
Colour privacy I included as it was one of the things hidden when a member was previously hidden, as for creation date I don't really remember why I added it but I believe people wanted it? I'd have to go back through chat logs to remember

PluralKit.Core/Migrations/6.sql
BeeFox-sys: This and the changelog for the api docs would need to be updated, i'll stick in a temporary date though

BeeFox-sys:
These updates bring this pr on level with current master, as well as merge some of the privacy settings (namely message count and creation date) into a new privacy option, `metadata` (if you can think of a better name I am happy to change it)  
It also includes the new last fronted and last message options in `pk;system list full` into the metadata privacy option.

Issue #37, opened 2018-12-23
Bot proxies server emojis as messages when a member has <> proxy tags


xSke:

spiralw:
Fixed in 94f4f970ea5828cd4469b3b44f8c9ad21f0cc960.

PR #175, opened 2020-05-14
Bug/proxy grt lst

BeeFox-sys:
This pull request solves issue #37 (_Bot proxies server emojis as messages when a member has <> proxy tags_), it is a simple fix and also works on proxy tags starting with <, and proxy tags ending with >.

### DIFF
--- a/PluralKit.Bot/Proxy/ProxyTagParser.cs
+++ b/PluralKit.Bot/Proxy/ProxyTagParser.cs
@@ -1,7 +1,7 @@
 ﻿#nullable enable
 using System.Collections.Generic;
 using System.Linq;
-
+using System.Text.RegularExpressions;
 using PluralKit.Core;
 
 namespace PluralKit.Bot
@@ -35,6 +35,12 @@ namespace PluralKit.Bot
                 
                 // Skip blank tags (shouldn't ever happen in practice)
                 if (tag.Prefix == null && tag.Suffix == null) continue;
+
+                var prefixPattern = new Regex(@"^<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>");
+                var suffixPattern = new Regex(@"<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>$");
+
+                if(tag.Prefix == "<" && prefixPattern.IsMatch(input)) continue;
+                if(tag.Suffix == ">" && suffixPattern.IsMatch(input)) continue;
 
                 // Can we match with these tags?
                 if (TryMatchTagsInner(input, tag, out result.Content))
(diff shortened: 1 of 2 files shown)


xSke:
What's with the docker-compose change?

BeeFox-sys:
The very first time I built it, it wouldn't build unless both the API and bot had build commands, the change allowed me to build it for the first time but hasn't been needed since (and needs to be removed in other branches I've submitted)

SophiaFoxyCoxy:
That won't fix animated emojis, nor will it fix mentions which also uses `<>`. You also won't fix all custom emojis as numbers and underscore can be used. The entire fix would be `^<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>(?:[\s\S]*>)?$`. However, I would do something else.

I suggest the following:
If prefix is `<`, test for `^<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>`.
If suffix is `>`, test for `<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>$`.

This should fix proxies `<text`, `text>`, and `<text>`.
For example:

Prefix: `<`
Suffix:
Stops proxying:
`<@0>`
`<@0> text`
Doesn't stop:
`< text<@0>`
`<<@0>`

Prefix:
Suffix: `>`
Stops proxying:
`<@0>`
`text <@0>`
Doesn't stop:
`<@0> text >`
`<@0>>`

Prefix: `<`
Suffix: `>`
Stops proxying:
`<@0>`
`text <@0>`
`<@0> text >`
`<<@0> text` - No suffix
Doesn't stop:
`<<@0>>`
`<text<@0>text>`

Explaination of the regex:
`^<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>`
`<(?:@!?|#|@&|a?:[\d\w_]+?:)\d+>$`
`^`: Is start of string
`<>`: are characters to match. These have no special properties.
`(?:)`: A group but doesn't store? Bascially you can't retrieve what's inside the group.
`|`: Basically an or. If the first set didn't match, try this.
`@!?`: Matches `<@0>` and `<@!0>`
`#`: Matches channel mentions
`@&`: Matches role mentions
`a?:[\d\w\_]+?:`: `a?` is to check if it's an animated emoji (`<a:name:0>`). `[\d\w_]+?` Matches all letters, numbers and underscore. `+?` Means 1 or more but end as soon as possible. So on the first `:` it sees, it will stop looking further instead of waiting for the last.
`\d+`: Is the id part of the mention or emoji.
`$`: Is end of string

spiralw:
Merged in 94f4f970ea5828cd4469b3b44f8c9ad21f0cc960.

PR #179, opened 2020-05-24
Various DSharpPlus workarounds

acw0:
I have not done extensive testing on these, however they all seem to work


acw0:
This should be it, let me know if I've missed anything

xSke:
I think I've essentially reimplemented this on my own (without really noticing), so I'm gonna close this for now. Thanks anyway, though~ :)

PR #182, opened 2020-06-08
Feature/prefix tags

BeeFox-sys:
This pr adds the feature to allow system tags to have a prefix and suffix, rather then just being a suffix to the name.
It also adds the ability to not have white space between the name and tag
and finally it maintains a legacy version where one does not need to define a prefix, or have the separator ("text" is the separator, in order to make it easier so users don't have to remember a new one)

tbh I don't expect this one to be merged but I thought it would be fun to have and create, and the pr exists in case you feel this way too

also ignore the "5 behind 7 ahead" it is actually only 2 ahead but git is weird)


spiralw:
This PR is quite outdated, and I'd like to implement it a little differently - either way, closing this, and I've added a note about it in to-do-list channel in the server.

PR #191, opened 2020-07-05
Return new member after member PATCH

greysdawn:
The current code returns the old member, resulting in a response with non-updated data; this pull request is meant to fix that


xSke:
Yep, my bad. Thanks~

PR #193, opened 2020-07-06
Find channel in all shards for `pk;msg`

acw0:
See commit messages.


acw0:
This makes an API call for every shard until we find the correct one. Thanks, DSharpPlus.
Closing this until I figure out how to do this properly.

PR #196, opened 2020-07-11
Fix URL handling when the filename contains multiple known extensions

acw0:
Example: `image.jpeg-32.jpg` gets matched as `image.jpeg`

I think this regex should work, although I wonder if there's anything else that can be after the filename in the URL other than a `?`


ent3r:
I think we should add `webp` and `gif` too, as those are supported as per the [discord API docs](https://discord.com/developers/docs/reference#image-data)

PR #197, opened 2020-07-19
Add git commit to help and stats commands

dev-kittens:


xSke:
Would it make more sense to try to query the `.git` directory at runtime instead?

```
$ cat .git/HEAD
ref: refs/heads/main
$ cat .git/refs/heads/main
176b5da485ab4f24b41a6e02d3f3e99c978ba701
$ cat .git/objects/17/6b5da485ab4f24b41a6e02d3f3e99c978ba701 | pigz -d
commit 250tree abcd1ad968d41cb6d75ba402e65d642744c6a333
parent 6054080dc76ae48831829f771dcde6ef370b0c05
author Ske <voltasalt@gmail.com> 1595084762 +0200
committer Ske <voltasalt@gmail.com> 1595084762 +0200

Properly block pinging normally unmentionable roles
```

Can extract commit hash and timestamp from that, without any CLI interfacing or external dependencies.

acw0:
Oh, sure. What file(s) should I put code that only gets run at init in?

xSke:
I'd probably put it in `PluralKit.Bot/Services/Something.cs`, have a function to fetch git info, then just cache it for further calls. Although there may be some edge cases, not sure if this is the best/most stable method. I just feel weird about build plugins for this kinda thing :p

acw0:
Hm. Would you prefer I run a shell command to decompress the git object, or include a zlib library in GitService?

xSke:
I think there's something in the standard library?

acw0:
Standard library seems to have gzip support, which I'm told is similar to zlib but needs a header prepended to it...? I can't figure out what I'm supposed to do for that.

xSke:
There's also this: https://github.com/libgit2/libgit2sharp

Figure if we're gonna bring in external dependencies, might as well bring in one that solves the problem properly instead of a hacky workaround. I'll look into it on my end, I think, though ;)

dev-kittens:
Closing as per the above comment. ^^

PR #200, opened 2020-07-21
Feature/default private membercards

BeeFox-sys:
Probably a little hacky but a neat and small modification to have member and system cards display only public information unless the flags `-all`, `-a`, or `-private` are used.


spiralw:
This is superseded (or will be, anyway) by #346.

PR #201, opened 2020-07-22
Fix renaming a member to a different capitalisation of the same name resulting in a confirmation message

acw0:


PluralKit.Bot/Commands/MemberEdit.cs
```diff
@@ -36,7 +36,8 @@ public MemberEdit(IDataStore data, IDatabase db)
 
             // Warn if there's already a member by this name
             var existingMember = await _data.GetMemberByName(ctx.System, newName);
-            if (existingMember != null) {
+            if (existingMember != null) 
+            if (existingMember.Id != target.Id) {
```
xSke, line 40: Should probably merge this to a && to prevent syntax confusion but I'll do that on my end :)

PR #202, opened 2020-07-22
Fix pk;msg errors across shards

acw0:
It appears the shards' guild cache *is* externally accessible - so we can just look up a guild there instead of using GetGuild (which would generate an API call per shard until we find the correct one).


xSke:
Have you tested that this works?

acw0:
we did, yeah

xSke:
ok :)

Issue #203, opened 2020-07-23
Migration script to upgrade from legacy to main


BakaKiller:
Hi there,
I can't find a script to migrate from a legacy version of the bot to the latter, "main" version
Is there anything anywhere ?

Edit : Not sure it matters but I use docker

Thanks for your hard work !

acw0:
Hi! The current bot includes migration code. If you run the latest code on a copy of your database, it should automatically perform all the required migrations.

xSke:
I'm not sure if the database schema changed between the legacy branch and the introduction of the migration system, so it *may* need a couple SQL commands to nudge things along. Try it and let me know?

BakaKiller:
So it does not look like it works. I mean, there are some migreations, but it seems like it is not enough :
```
[2020-07-23 20:09:53.772440] FTL Error while running bot
Npgsql.PostgresException (0x80004005): 42703: column guild.blacklist does not exist
   at Npgsql.NpgsqlConnector.<>c__DisplayClass160_0.<<DoReadMessage>g__ReadMessageLong|0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Npgsql.NpgsqlConnector.<>c__DisplayClass160_0.<<DoReadMessage>g__ReadMessageLong|0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming)
   at Npgsql.NpgsqlCommand.ExecuteReaderAsync(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at PluralKit.Core.PKCommand.LogQuery[T](Task`1 task) in /app/PluralKit.Core/Database/Wrappers/PKCommand.cs:line 106
   at Dapper.SqlMapper.ExecuteImplAsync(IDbConnection cnn, CommandDefinition command, Object param) in C:\projects\dapper\Dapper\SqlMapper.Async.cs:line 678
   at PluralKit.Core.Database.ExecuteSqlFile(String resourceName, IPKConnection conn, IDbTransaction tx) in /app/PluralKit.Core/Database/Database.cs:line 126
[...]

PR #204, opened 2020-07-24
Add upgrade from legacy doc in README (#203)

BakaKiller:


xSke:
I remember we migrated the production table using a single query, not a generated one; shouldn't be too hard to make it do that?

Something like:
```sql
update members set proxy_tags = array((prefix, suffix))
```

although I'm almost entirely sure that's the wrong syntax. Would need some playing around. I can *never* remember how Postgres does arrays... which probably means we should migrate away from them to a proper relational model at some point. But that's less fun.

BakaKiller:
From what i've just checked, it should be something like `proxy_tags = {(prefix, suffix)}`, i'll make a few tests and edit that !

xSke:
This should work:
```sql
update members set proxy_tags = array[(members.prefix, members.suffix)]::proxy_tag[] where members.prefix is not null or members.suffix is not null;
```

see: http://sqlfiddle.com/#!17/6ddd3

PR #208, opened 2020-07-29
Add configurable latch timeout

dev-kittens:
I'm not 100% on this one, but putting it out there either way. 
Info strings probably need updated as well.


dev-kittens:
will probably do this elsewhere

PR #213, opened 2020-08-12
Fix doubleBacktick replacing

dev-kittens:
the first run of doubleBacktick.replace has a visible space instead of a zero width space. this copy pastes the second run's replace string into the first run so the visible output is unchanged.


xSke:
Can we use the unicode escape sequence instead, just to alleviate confusion?On 12 Aug 2020 16:55, kittens <notifications@github.com> wrote:the first run of doubleBacktick.replace has a visible space instead of a zero width space. this copy pastes the second run's replace string into the first run so the visible output is unchanged.



You can view, comment on, or merge this pull request online at:
  https://github.com/xSke/PluralKit/pull/213

Commit Summary
Fix doubleBacktick replacing

File Changes

    M
    PluralKit.Bot/Utils/DiscordUtils.cs
    (4)
  

Patch Links:
https://github.com/xSke/PluralKit/pull/213.patchhttps://github.com/xSke/PluralKit/pull/213.diff

dev-kittens:
Good idea. Just pushed that

PR #215, opened 2020-08-21
Add flags to switch delete all

dev-kittens:


xSke:
Good catch :) Might want to remove the "subcommand" modes at some point but this is good for now.

PR #229, opened 2020-10-09
Use browser prefers-color-scheme for dark theme docs

dev-kittens:


xSke:
Does this also have a way to manually toggle it on/off?

dev-kittens:
It doesn't. I'll look into that.

dev-kittens:
> Does this also have a way to manually toggle it on/off?

@xSke we looked at that a while ago, and it seems that Vuepress doesn't have an easy way to do that, unfortunately.

PR #233, opened 2020-10-18
Allow deleting bot responses with :x: reaction

### LINKED ISSUE #170: Allow deleting bot responses with an X reaction
- This would only work if done by the account that originally "triggered" the response
- Useful for privacy, deleting responses/cards potentially containing sensitive information
- Needs to "work around" the existing X reaction on the scrolling cards, etc (or just repurpose that to make it delete the card entirely)
- Would have to save a mapping of bot response message IDs -> "trigger accounts", likely only stored for a couple of minutes each
  - either through IMemoryCache or a plain map that gets "garbage collected" by a scheduled task

dev-kittens:
Closes #170.

I chose to put this in the database instead of a bot cache because it would be unfortunate if someone posted some information they then wanted to delete, and the bot was restarted right after.

Currently, only messages that contain an embed are able to be deleted by reaction.
The timeout is 1 hour, and old rows get cleaned up by the once-a-minute scheduled task.

### DIFF
--- a/PluralKit.Bot/Commands/CommandTree.cs
+++ b/PluralKit.Bot/Commands/CommandTree.cs
@@ -189,9 +189,9 @@ namespace PluralKit.Bot
             if (ctx.Match("random", "r"))
                 return ctx.Execute<Member>(MemberRandom, m => m.MemberRandom(ctx));
 
-            ctx.Reply(
+            // remove compiler warning
+            return ctx.Reply(
                 $"{Emojis.Error} Unknown command {ctx.PeekArgument().AsCode()}. For a list of possible commands, see <https://pluralkit.me/commands>.");
-            return Task.CompletedTask;
         }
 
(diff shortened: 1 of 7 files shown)


xSke:
I'd probably do this just as an inline SQL instead of a DB function. Avoids dealing with too many migrations. But I'll merge and make that change on my end :) Thanks a lot!

PluralKit.Bot/Commands/CommandTree.cs
```diff
@@ -189,9 +189,9 @@ public Task ExecuteCommand(Context ctx)
             if (ctx.Match("random", "r"))
                 return ctx.Execute<Member>(MemberRandom, m => m.MemberRandom(ctx));
 
-            ctx.Reply(
+            // remove compiler warning
```
xSke, line 192: Good catch, dunno how this snuck in there :)

PR #234, opened 2020-10-19
Add "member is public" reminder message on member creation

dev-kittens:
(if system has any private members - that is, any members with visibility set to private)


xSke:
I think here it might be better to just query `select count(*) from members where system = @System and member_visibility = 2`. Might already have a helper for something like that?

PR #264, opened 2020-12-12
Add proxy debug command

spiralw:
A note: the message with the `pk;proxy debug` command does need to be a message that would be otherwise proxied; so it needs to include suffix proxy tags or for the author to have `pk;` as a proxy tag. 
Would it be possible to create a system with an example member with `pk;` proxy tag, and give that system ID to `GetProxyMembers` when debugging?


spiralw:
This is not a great implementation, so I'm closing this, but I've added a note in the to-do list on the Discord server.

PR #269, opened 2020-12-22
Port to custom Discord library

xSke:


silasary:
Why is this needed?

xSke:
DSharpPlus breaks. like, all the time, in increasingly funky and impossible-to-debug ways. Plus, they're pretty slow to implement new features and I'd rather not wait three months to add proxy message editing.

silasary:
Yeah, makes sense

PR #273, opened 2021-01-13
[doc] Clarifying <group>/<name> into <groupname> in command-list.md

FirePowi:
Open for discussion

### DIFF
--- a/docs/content/command-list.md
+++ b/docs/content/command-list.md
@@ -57,20 +57,20 @@ Words in **\<angle brackets>** or **[square brackets]** mean fill-in-the-blank. 
 - `pk;member <name> delete` - Deletes a member.
 
 ## Group commands
-*Replace `<name>` with a group's name, 5-character ID or display name. For most commands, adding `-clear` will clear/delete the field.*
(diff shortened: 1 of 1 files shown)


FirePowi:
Can have help with checks, please? I don't understand what's wrong?

dev-kittens:
> Can have help with checks, please? I don't understand what's wrong?

GitHub builds are currently broken. You're only updating documentation, so don't worry about it.

FirePowi:
@dev-kittens Thanks!

xSke:
I've been using `<group>` similar to `<member>`/etc to imply that you can both use the name or the ID. I feel like `<groupname>` would muddle that up, but also there might be an even better way to clarify that somehow?

FirePowi:
@xSke I think `<member>` and `<group>` is better, that just that it was `<name>` and `<name>` is ambiguous for 
`pk;member new <name>`/`pk;group new <name>` as in this command, you can't give the ID, but only a name to give to the new member/group

spiralw:
Merged in 28cd50a5bfc4be61b8f15a9545b0b21511f3bef1.

PR #274, opened 2021-01-13
Add display options to short member list

spiralw:
This PR adds showing the following to the short member list as properties 
- Birthday
- Message count
- Last front
- Creation date
- Pronouns (only shown, not sorted), via a new `with-pronouns` flag
- Last proxy timestamp

This also adapts the full list to show the sort property's relevant item without needing to use *both* the sort and show flag.


xSke:
Might look neater to put it in parentheses instead of with colons? eg. 

> [id] Member Name (123 messages)

spiralw:
Merged in 2807906d7904619d1505a320433ef8e15acafbb9.

PR #295, opened 2021-03-28
Add system and group colors

Spectralitree:
Basically what it says on the tin.

- Add a color column to both the systems and groups models.
- Add two commands that handle setting the colors
- Change the embed services so the color actually shows up on the system and group embeds

EDIT: added two more things!

- The system and group embeds have a textfield listing the color now
- The member list, group list, and group member list all have colors now.

Frontpercent does not have the color as of yet.

image
(yay, colors!)

image
(even more color!)

### DIFF
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -9,8 +9,6 @@ using Dapper;
 
 using Humanizer;
 
-using NodaTime;
-
 using Myriad.Builders;
 
 using PluralKit.Core;
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -311,7 +311,7 @@ namespace PluralKit.Bot
             }
 
             var title = system.Name != null ? $"Groups of {system.Name} (`{system.Hid}`)" : $"Groups of `{system.Hid}`";
-            await ctx.Paginate(groups.ToAsyncEnumerable(), groups.Count, 25, title, Renderer);
+            await ctx.Paginate(groups.ToAsyncEnumerable(), groups.Count, 25, title, ctx.System.Color, Renderer);
             
             Task Renderer(EmbedBuilder eb, IEnumerable<ListedGroup> page)
             {
@@ -390,7 +390,7 @@ namespace PluralKit.Bot
             if (opts.Search != null) 
                 title.Append($" matching **{opts.Search}**");
             
-            await ctx.RenderMemberList(ctx.LookupContextFor(target.System), _db, target.System, title.ToString(), opts);
+            await ctx.RenderMemberList(ctx.LookupContextFor(target.System), _db, target.System, title.ToString(), target.Color, opts);
         }
 
         public enum AddRemoveOperation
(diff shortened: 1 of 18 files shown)


PluralKit.Bot/Commands/Groups.cs
```diff
@@ -224,6 +225,53 @@ async Task ShowIcon()
...
+                        .Thumbnail(new($"https://fakeimg.pl/256x256/{target.Color}/?text=%20"))
+                        .Description($"This group's color is **#{target.Color}**."
+                                         + (ctx.System?.Id == target.System ? $" To clear it, type `pk;group {target.Reference()} color -clear`." : ""))
+                        .Build());
+            }
+            else
+            {
+                ctx.CheckOwnGroup(target);
+
+                if (color.StartsWith("#")) color = color.Substring(1);
```
xSke, line 262: I think we have a utility function for color parsing like this?

PluralKit.Bot/Commands/SystemEdit.cs
```diff
@@ -91,6 +92,47 @@ await ctx.Reply(embed: new EmbedBuilder()
...
+                        .Thumbnail(new($"https://fakeimg.pl/256x256/{ctx.System.Color}/?text=%20"))
+                        .Description($"Your system's color is **#{ctx.System.Color}**. To clear it, type `pk;s color -clear`.")
+                        .Build());
+            }
+            else
+            {
+                var color = ctx.RemainderOrNull();
+
+                if (color.StartsWith("#")) color = color.Substring(1);
+                if (!Regex.IsMatch(color, "^[0-9a-fA-F]{6}$")) throw Errors.InvalidColorError(color);
```
xSke, line 124: If there is a utility function, should be used here and the other place :)

PluralKit.Bot/Commands/SystemEdit.cs
```diff
@@ -91,6 +92,47 @@ await ctx.Reply(embed: new EmbedBuilder()
...
+            if (await ctx.MatchClear())
+            {
+                var patch = new SystemPatch {Color = Partial<string>.Null()};
+                await _db.Execute(conn => _repo.UpdateSystem(conn, ctx.System.Id, patch));
+
+                await ctx.Reply($"{Emojis.Success} System color cleared.");
+            }
+            else if (!ctx.HasNext()) 
+            {
+                if (ctx.System.Color == null)
```
xSke, line 108: Does/should this allow you to look up other systems' colors?

xSke (approved):
Looks good, thank you for this!!

PluralKit.Bot/Commands/Groups.cs
Spectralitree: If we do, it's not used in the member color command either! I just mostly referenced/copied over the code from the member command

PluralKit.Bot/Commands/SystemEdit.cs
Spectralitree: There's not a command for specifically looking up a system's color by specifying a target. But pulling up the system card in general shows the color and hex code field like it would on the member card.

xSke:
Sorry for the delay. thanks again!

PR #297, opened 2021-03-31
fix channel checks missing GuildNews

spiralw:
I left the log channel / log clean checks as-is because i'm assuming nobody's using an announcements channel as their log channel (I hope).


spiralw:
Merged in e620e30c10d8fc7c0dbf0221991a2283ebf54b87.

PR #302, opened 2021-04-02
Autoproxy scopes

spiralw:
This PR adds support for autoproxy scopes (global/server/channel autoproxy).

It works, but it still needs a lot of work, namely:
- properly showing autoproxy status ("effective settings" vs actual settings for chosen scope)
- migrating existing autoproxy configs
- latch timeout
- selecting a channel or server instead of just acting on the *current* channel or server
- there's some not-great SQL to be refactored
- documentation, documentation, documentation!
- ... and more!


spiralw:
will do this somewhere else later

PR #303, opened 2021-04-05
Add banner (large) image

spiralw:


spiralw:
a thought: should this be added to the API? (how would that even be versioned?)

Spectralitree:
> a thought: should this be added to the API? (how would that even be versioned?)

Definitely would appreciate it if this got added to the API, we already got a couple of ideas of stuff we want to do with it!

Not sure how to go about versioning either, tho

spiralw:
Merged in e1445719043d2e728fd9cde0c6e3518ebb454b6a.

PR #307, opened 2021-04-06
Add color to frontpercent embed

Spectralitree:
Title is self explanatory!


spiralw:
just a quick note: you're passing `target` to `CreateFrontPercentEmbed` already, there's no need to also *separately* pass the color

Spectralitree:
oops, how did I not notice that, I'll change that when i get the chance!

PR #308, opened 2021-04-08
Add message editing

### LINKED ISSUE #283: [Feature Request] Editing Messages
You should be able to edit proxied messages, instead of deleting and resending them. This could be done with a PATCH request (https://discord.com/developers/docs/resources/webhook#edit-webhook-message). Maybe have a reaction, then the next message after that is the new text? Anyway, I'm going to write a JS bot to do this,  but as I don't know Go, it could be good for you to implement similar code, based on my code. I'll link the github repo when I finish!

Qyriad:
`pk;message <message> edit <new message>`, or `pk;edit <message> <new message>` for short, where `<message>` can be an ID or link, or can be omitted entirely if the command is sent as a reply to a message. This also makes the existing `pk;message` functionality accept a reply in place of a link or ID.

Closes #283.

### DIFF
--- a/Myriad/Rest/DiscordApiClient.cs
+++ b/Myriad/Rest/DiscordApiClient.cs
@@ -121,6 +121,10 @@ namespace Myriad.Rest
             _client.PostMultipart<Message>($"/webhooks/{webhookId}/{webhookToken}?wait=true",
                 ("ExecuteWebhook", webhookId), request, files)!;
 
+        public Task<Message> EditWebhookMessage(ulong webhookId, string webhookToken, ulong messageId, EditWebhookMessageRequest request) =>
+            _client.Patch<Message>($"/webhooks/{webhookId}/{webhookToken}/messages/{messageId}",
+                ("EditWebhookMessage", webhookId), request)!;
+
         public Task<Channel> CreateDm(ulong recipientId) =>
             _client.Post<Channel>($"/users/@me/channels", ("CreateDM", default), new CreateDmRequest(recipientId))!;
 
--- a/Myriad/Rest/Types/Requests/EditWebhookMessageRequest.cs
+++ b/Myriad/Rest/Types/Requests/EditWebhookMessageRequest.cs
@@ -1,0 +1,11 @@
+using Myriad.Types;
+
+namespace Myriad.Rest.Types.Requests
+{
+    public record EditWebhookMessageRequest
+    {
+        public string Content { get; init; }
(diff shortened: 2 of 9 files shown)


spiralw:
This breaks `pk;msg [id|link]` functionality.
- `ctx.PopArgument()` returns an empty string if it finds no argument, and an empty string is not null, so the "you must [...]" will never get thrown.
- `PrintCommandExpectedError` is not the right thing to throw if the message was not found in the database. It is `Errors.MessageNotFound(messageId)`.
- you didn't break the link matching regex ([I did](https://github.com/xSke/PluralKit/commit/a5ad16780c2bd9b9a742727b5c3f3a4817b51aa1)) but this is still missing parsing ulong *before* checking the regex, so it won't work with just a message ID.

Some other notes:
- `pk;edit` raw command also only works with replies, but as it doesn't log thrown errors anywhere, I don't know why. (any `PKError`s thrown before `ctx.Execute` don't get sent to the channel, so you would need to do something [like this](https://github.com/xSke/PluralKit/blob/main/PluralKit.Bot/Commands/CommandTree.cs#L490-L494))
- You need to add an error message for when the bot doesn't have manage webhook permissions, and for when the original webhook was deleted (can't edit the message with a different webhook).
- If the edit command message was sent in a different channel (or in DMs), don't delete it, instead send a message saying the message was edited successfully.

Qyriad:
> You need to add an error message for when the bot doesn't have manage webhook permissions
Unless I'm mistaken, it doesn't seem that editing webhook messages actually requires the manage webhook permission. Do you still want it to error without that permission?

spiralw:
It does need that permission to fetch the webhook. Try disabling the permission and then restarting the bot (to clear its internal webhook cache).

spiralw:
From the pluralkit server: 

> Wouldn't it be easier to be able to edit the message with an emoji reaction like both PK Edit and [Tupperbox] do?

Qyriad:
> Wouldn't it be easier to be able to edit the message with an emoji reaction like both PK Edit and [Tupperbox] do?

Tupperbox's `tul!edit` is very similar to this, but sure I can do the emoji reaction too

Qyriad:
We were literally just about to push our updated changes to this, but it looks like y'all already have that covered with 3d624b39e4240be69232da2f54adb465abf69ebb? Should we close this PR then?

xSke:
Yes, this is obsoleted. Sorry about that.

PR #309, opened 2021-04-09
Fix matching Discord Stable links

spiralw:


xSke:
damn, how did I miss that during review... embarassing lol

PR #310, opened 2021-04-12
Remove hardcoded prefixes in strings

spiralw:
this allows for, in the future, changing pluralkit's prefix per-server, i haven't done that here though


xSke:
At this point I'd rather save this for a larger refactor in how string generation is handled.

spiralw:
yeah, that's fair

PR #313, opened 2021-04-14
feat: add avatar to member via attachment on 'pk;member new' command

spiralw:


xSke:
not entiiiiirely happy with how the error handling is done here but it should be fine for now

PR #316, opened 2021-04-21
API patch improvements

spiralw:
- add PatchObject.CheckIsValid
- use transaction when creating member, as to not create a member if the
patch is invalid
- return edited system in `PATCH /s` endpoint


xSke:
Good stuff. Should hopefully fix the broken avatar errors.

PR #323, opened 2021-05-01
misc patches

spiralw:
- feat: case-insensitive "text" keyword matching for proxy tags
- fix: don't try matching commands if there is nothing to match
- feat: show member color in reply embed

### DIFF
--- a/PluralKit.Bot/Commands/MemberProxy.cs
+++ b/PluralKit.Bot/Commands/MemberProxy.cs
@@ -26,6 +26,7 @@ namespace PluralKit.Bot
             {
                 // // Make sure there's one and only one instance of "text" in the example proxy given
                 var prefixAndSuffix = exampleProxy.Split("text");
+                if (prefixAndSuffix.Length == 1) prefixAndSuffix = prefixAndSuffix[0].Split("TEXT");
                 if (prefixAndSuffix.Length < 2) throw Errors.ProxyMustHaveText;
                 if (prefixAndSuffix.Length > 2) throw Errors.ProxyMultipleText;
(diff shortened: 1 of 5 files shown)


PluralKit.Bot/Commands/MemberProxy.cs
```diff
@@ -26,6 +26,7 @@ ProxyTag ParseProxyTags(string exampleProxy)
             {
                 // // Make sure there's one and only one instance of "text" in the example proxy given
                 var prefixAndSuffix = exampleProxy.Split("text");
+                if (prefixAndSuffix.Length == 1) prefixAndSuffix = prefixAndSuffix[0].Split("TEXT");
```
xSke, line 29: hmm, is there a way to do this "nicer"? like a case-insensitive split?

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -194,11 +194,14 @@ private Embed CreateReplyEmbed(Message trigger, Message repliedTo, string? nickn
             var username = nickname ?? repliedTo.Author.Username;
             var avatarUrl = $"https://cdn.discordapp.com/avatars/{repliedTo.Author.Id}/{repliedTo.Author.Avatar}.png";
 
+            Console.WriteLine($"color {match.Member.Color}");
```
xSke, line 197: remove this

PluralKit.Bot/Commands/MemberProxy.cs
spiralw: could use a regex instead, but that seems a bit overkill?

xSke: yeah, fair.

PR #328, opened 2021-05-08
Minor change to user-guide.md

rladenson:
It said that pk doesn't have some things that have since been implemented.


rladenson:
I don't know why it says line 614 was changed I didn't change it and nothing looks changed

PR #337, opened 2021-05-18
message logging tweaks

spiralw:
- use name/avatar from sent message instead of fetching it from `LookupContext.ByNonOwner` (which might be different from the actual proxied message)
- deduplicate code between normal logging and edited message logging


spiralw:
Merged in 9fd4f327e0f199d4f33f439483bc50bced585b0e.

PR #340, opened 2021-05-19
Add server-specific system tag

### LINKED ISSUE #281: Server-specific system tags
This makes a lot of sense on e. g. servers around a particular language, and translating the tags for them.

spiralw:
Closes #281.

h


BeautifulPixel:
Would it make sense to support setting the server tag as the empty string or otherwise disabling tag display on a specific server?

spiralw:
whoops, forgot about that. will add that later today

spiralw:
Merged in 9d4006b05648d06d59ee09a9226f3b0c725b2c5f.

PR #341, opened 2021-05-24
Tweak -raw usage

Spectralitree:
This PR does a couple of things:

- MatchRaw in ContextArgumentExt checks for -r and -raw with and without the dash now.
- All the commands that checked for a -raw flag now use MatchRaw instead
- Some more fields like the system tag are able to be printed using -raw now

### DIFF
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -101,32 +101,30 @@ namespace PluralKit.Bot
 
                 await ctx.Reply($"{Emojis.Success} Group display name cleared.");
             }
+            else if (ctx.MatchRaw())
+            {
+                // No perms check, display name isn't covered by member privacy
+                if (target.DisplayName == null)
+                {
+                    if (ctx.System?.Id == target.System)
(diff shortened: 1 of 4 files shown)


PluralKit.Bot/Commands/Groups.cs
```diff
@@ -150,16 +148,29 @@ public async Task GroupDescription(Context ctx, PKGroup target)
...
+                if (target.Description == null)
+                    if (ctx.System?.Id == target.System)
+                        await ctx.Reply($"This group does not have a description set. To set one, type `pk;group {target.Reference()} description <description>`.");
+                    else
+                        await ctx.Reply("This group does not have a description set.");
+                else await ctx.Reply($"```\n{target.Description}\n```");
             } 
             else if (!ctx.HasNext())
             {
+                if (!target.DescriptionPrivacy.CanAccess(ctx.LookupContextFor(target.System)))
```
spiralw, line 166: Can you move this change to #342?

Spectralitree: realized that just now as well. give me a second

spiralw:
Merged in 0553e42eca643544db9bb220839493610e633e50.

In the future, please don't duplicate strings - it's harder to maintain that way

PR #346, opened 2021-05-29
Do not show private fields in system card unless passed -all

Qyriad:


xSke:
Should this extend to other cards? member card, group card, etc

Qyriad:
Probably, yes, but we only implemented it for system cards in this PR. We can do the others too in another PR if you'd like.

xSke:
This is a change that IMO should be properly announced and "telegraphed" so it would be nice to have it happen all at once

Qyriad:
Gotcha, we'll do that in a bit, then 👍

spiralw:
This branch is quite outdated and I want to generalize the idea a bit more: #404

PR #353, opened 2021-06-09
Import/export improvements

### LINKED ISSUE #277: Inconsistent proxy tag length limit
When importing, there is a proxy tag length limit of 100 characters. This limit isn't present when actually registering proxy tags through the bot, which ends up with users not being able to export and re-import because one of their members has proxy tags longer than 100 characters.

spiralw:
- import/export groups
- import/export privacy settings
- cleaned up BulkImporter code (now uses Patch.Apply instead of duplicating code)
- removed `DataFileMember` and added JSON properties to `PKMember` / `PKGroup` (so new member props are immediately added to the export file)

There's a couple things I'd like to add (like importing Tupperbox groups) but I'll work on that at some other time.

Also closes #277.

### DIFF
--- a/PluralKit.Bot/Commands/ImportExport.cs
+++ b/PluralKit.Bot/Commands/ImportExport.cs
@@ -89,7 +89,10 @@ namespace PluralKit.Bot
                     else
                     {
                         // We already had a system, so show them what changed
-                        await ctx.Reply($"{Emojis.Success} Updated {result.ModifiedNames.Count} members, created {result.AddedNames.Count} members. Type `pk;system list` to check!");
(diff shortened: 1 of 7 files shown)


spiralw:
On second thought, it might be a good idea to grab the JSON serialization stuff in the API and put *that* here as well (is there anything different between what that generates and what needs to go in an export file?)

xSke:
I was considering "unifying" that somehow? or at least make the formats as similar as possible. might be possible to do some code sharing there

spiralw:
One issue I've just found is the API serializes timezone as `tz` and export files serialize it as `timezone`; other than that, they seem to be fully compatible

xSke:
Yeah, things like that would be nice to clean up for an export format v2 along with api v2

PR #361, opened 2021-06-21
Add /v1/meta API endpoint

spiralw:
Shows shard state (id, up/down, ping duration, last heartbeat time, last connection time)

Is there anything else that should be included here?


xSke:
I'd actually remove the count estimates entirely? Would rather expose no data than "bad" data.

spiralw:
ah, I guess we misinterpreted our conversation a while ago about using estimates, then
sure, can do, those stats are available via API by fetching from grafana anyway

xSke:
Yeah; I'd be fine if they were close estimates but I think the message count was tens of millions off? so, not great :/

PR #364, opened 2021-06-25
Fix mentions in replies being cut off

zoemartin01:
This fixes discord mentions such as emotes or channel mentions being cut off in the middle due to the 100 character limit when replying via proxy. 

Before:
before

After:
after


zoemartin01:
Also added the same thing for URLs being cut off

PR #367, opened 2021-06-30
Avoid killing the gateway session when trying to resume

spiralw:
Sending a 1000 close code invalidates the session, which means a shard will go offline for a few seconds when it needs to reconnect. Instead, we can use a 1005 close code (we could use any other non-1000 or 1001 code but 1005 seems more *correct*).


spiralw:
Merged in 889e1a83315563b1d3cb916c8600596d37dfe29a.

PR #374, opened 2021-07-18
Message info embed improvements

starshine-sys:
- Fetch the user if there's no member object, so users who left the server don't show up as deleted
- Order roles by position (so potential staff roles show up first and aren't cut off)

### DIFF
--- a/PluralKit.Bot/Services/EmbedService.cs
+++ b/PluralKit.Bot/Services/EmbedService.cs
@@ -301,6 +301,8 @@ namespace PluralKit.Bot {
                 if (member != null)
                     // Don't do an extra request if we already have this info from the member lookup
                     userInfo = member.User;
+                else userInfo = await _cache.GetOrFetchUser(_rest, msg.Message.Sender);
+
                 memberInfo = member;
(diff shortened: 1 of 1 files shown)


PluralKit.Bot/Services/EmbedService.cs
```diff
@@ -301,6 +301,8 @@ public async Task<Embed> CreateMessageInfoEmbed(FullMessage msg)
                 if (member != null)
                     // Don't do an extra request if we already have this info from the member lookup
                     userInfo = member.User;
+                else userInfo = await _cache.GetOrFetchUser(_rest, msg.Message.Sender);
+
                 memberInfo = member;
             }
             else userInfo = await _cache.GetOrFetchUser(_rest, msg.Message.Sender);
```
spiralw, line 308: as to not duplicate code, would it be possible to `if (userInfo == null) userInfo = await ...` outside of the `if (channel != null)` block?

spiralw:
Merged in 74140b59944ee3ec12a7e144027de14c92f68903.

PR #380, opened 2021-08-03
Fix Discord merging webhook messages with same username

spiralw:
This works, but I'm not sure if there's any issues that might pop up in channels with lots of activity.

The actual checking if it needs to fix the name is simple enough; finding the last proxied message is a larger issue:
- a message can be deleted, but still be in the LastMessageCache. So, I added a method that listens to MessageDelete event and cleans up the cache.
- the LastMessageCache can be missing or out of date (too many messages were deleted, or the bot was restarted). So, we fetch the last message from Discord.


spiralw:
Merged in 25f96dd920b9a1de187fc04640581f16245289b6.

PR #382, opened 2021-08-03
Feature proxy permcheck

Spectralitree:
This PR adds a `pk;proxycheck` command, which goes through all the checks for a non-proxied message and sends a message detailing why a message hasn't been proxied.

It works by changing how the proxy checks work, throwing a ProxyChecksFailedException instead of silently failing. This exception gets ignored by the global exception handler. `pk;proxycheck` fetches the message using a reply or a message link, then runs it through the same proxy checks, and replies with the ProxyChecksFailedException message.

### DIFF
--- a/PluralKit.Bot/Commands/Misc.cs
+++ b/PluralKit.Bot/Commands/Misc.cs
@@ -17,6 +17,7 @@ using Myriad.Cache;
 using Myriad.Extensions;
 using Myriad.Gateway;
 using Myriad.Rest;
+using Myriad.Rest.Exceptions;
 using Myriad.Rest.Types.Requests;
 using Myriad.Types;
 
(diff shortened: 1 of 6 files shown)


PluralKit.Bot/Commands/Misc.cs
```diff
@@ -247,5 +252,79 @@ public async Task GetMessage(Context ctx)
...
+
+            // check if the message is not a webhook and sent by the same user as the message that's being checked
+            if (msg.WebhookId != null)
+                throw new PKError("You cannot check messages sent by a webhook.");
+            if (msg.Author.Id != ctx.Author.Id)
+                throw new PKError("You can only check your own messages.");
+
+
+            // get the channel info
+            var channel = await _rest.GetChannel(msg.ChannelId);
```
spiralw, line 303: This should be getting the channel from cache, no need for an API request here (just fail if it's not found)

spiralw:
looks good, but your base branch is really old and missing some important fixes. please fetch latest `dev` branch and rewrite your commits on top of that

spiralw:
nice! a couple things missing, but I just added those myself :)

Merged in b9f73cadb74949d33b9224ff0e87f60be01897dc.

PR #386, opened 2021-08-21
Add @/everyone perm checks to pk;permcheck

Alyx-W:


PluralKit.Bot/Commands/Misc.cs
```diff
@@ -214,9 +228,29 @@ public async Task PermCheckGuild(Context ctx)
                     eb.Color(DiscordUtils.Red);
                 }
             }
+            string footer = "";
+
+            foreach (var (missingPermissionField, channels) in permissionsMissing)
```
spiralw, line 233: please use LINQ :p

PluralKit.Bot/Commands/Misc.cs
```diff
@@ -214,9 +228,29 @@ public async Task PermCheckGuild(Context ctx)
...
             if (hiddenChannels > 0)
-                eb.Footer(new($"{"channel".ToQuantity(hiddenChannels)} were ignored as you do not have view access to them."));
+            {   
+                if (footer.Length > 0)
+                {
+                    footer = $"{"channel".ToQuantity(hiddenChannels)} were ignored as you do not have view access to them.";
+                }
+                else
+                {
+                    footer += $"\n{"channel".ToQuantity(hiddenChannels)} were ignored as you do not have view access to them.";
```
spiralw, line 249: I don't think you can use newlines in an embed footer? either way, this should not be duplicated - maybe do something like
```cs
if (footer.Length > 0)
    footer += "\n";

footer += $"[...]";
```

spiralw (approved):

PluralKit.Bot/Commands/Misc.cs
```diff
@@ -214,9 +228,29 @@ public async Task PermCheckGuild(Context ctx)
...
+                }
```
xSke, line 250: I wouldn't leak the amount of hidden channels, actually?

I know this code was here already but it was never a good idea in the first place :)

spiralw: does this message need to be here at all? I can only think of it being useful if the bot's cache is broken and it thinks the user is missing permissions when they actually aren't

PluralKit.Bot/Commands/Misc.cs
Alyx-W: > please use LINQ instead, it's a lot cleaner

Maybe respond in #internal-stuff, but definitely, I just wasn't sure how, I would really love a quick example if you have time of how to do that better, I haven't worked with LINQ much yet

PluralKit.Bot/Commands/Misc.cs
xSke: Yeah, maybe not? We probably don't lose anything from just removing it outright

PluralKit.Bot/Commands/Misc.cs
spiralw: I haven't tried this, but it should work:
```cs
if (permissionsMissing.Any((missingPermissionField, _) => missingPermissionField == (ulong)PermissionSet.UseExternalEmojis)
```

PR #389, opened 2021-08-31
Put "Queried by" info in footer for deletion of older command responses

rladenson:


spiralw:
as per #polls, this PR is superseded by extending the deletion period to 1 day and by allowing any messages to be deleted in DMs.

PR #403, opened 2021-11-18
fix(api): patch correct system privacy fields

Spectralitree:
patching `privacy.member_list_privacy` and others with the same issue should now update their respective fields instead of the description


spiralw (approved):
this is what I get for copy-pasting

PR #414, opened 2022-01-17
Added -with-displayname flag to member and group list

rladenson:
Even if dn gets added back to group list as a default this flag should probably still exist for member list


spiralw:
Merged in c62b9c2d4064672c0d9b1654b5fbf72942d47041

PR #428, opened 2022-03-17
Fixed the messages about clearing the member and system color

UsernameNotAvalible:
This is a pretty simple change. When viewing the color of your own system (if it is set), a message will display saying to run 'Pk;s color -clear' to clear the system color. However, 'Pk;s color -clear' is actually incorrect usage, so this change corrects the error


spiralw:
I'd prefer if this actually fixed the `-clear` flag instead of just hiding it.

PR #429, opened 2022-03-23
system pronouns

Draconizations:
adds a pronouns field to the system card, and a system pronoun privacy setting (since members also have a pronouns privacy setting)


spiralw:
looks good! one quick nitpick: could you add the new JSON keys to the API models?

PR #430, opened 2022-03-23
De-Latch command

UsernameNotAvalible:
This adds a de-latch command, which will "forget" the last-proxied member so that new messages will not be autoproxied until another member is manually proxied. The command is `Pk;ap delatch`


spiralw:
unlatching is already possible by starting a message with two backslashes (`\\`). is a separate command needed for that?

PR #435, opened 2022-03-31
Improves the proxy command

UsernameNotAvalible:
This change slightly improves the `Pk;m name proxy` command so that trying to set a new proxy without the word "text" (such as `Pk;m TestGuy proxy TT:`) will assume that the new proxy is a prefix (So `Pk;m TestGuy proxy TT:` acts the same as `Pk;m TestGuy proxy TT:text`)


spiralw:
hey, please talk to us in the support server (or at least open an issue) before sending PRs. this specifically is not a change i'm really interested in making

also please clean up your git branches in the future; I can squish your changes for the existing PRs but would be nice not to have to do that.

PR #436, opened 2022-03-31
Add prepend and append switches to Pk;edit

UsernameNotAvalible:
This change adds the flags `-prepend` and `-append` to the `Pk;edit` command. These will prepend or append the given text to the old message, rather than overwriting it completely. Example: If you have a message that says "test", you could use the command `Pk;edit -prepend This is a` to change the message contents to "This is a test". Then, you could use `Pk;edit -append message that is cool` to change the message to "This is a test message that is cool"

### DIFF
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -51,14 +51,26 @@ public class ProxiedMessage
         if (!ctx.HasNext())
             throw new PKSyntaxError("You need to include the message to edit in.");
 
-        var newContent = ctx.RemainderOrNull().NormalizeLineEndSpacing();
-
-        if (newContent.Length > 2000)
-            throw new PKError("PluralKit cannot proxy messages over 2000 characters in length.");
-
         var originalMsg = await _rest.GetMessageOrNull(msg.Message.Channel, msg.Message.Mid);
         if (originalMsg == null)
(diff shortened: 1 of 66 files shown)


PluralKit.Bot/Commands/Message.cs
```diff
@@ -51,15 +51,27 @@ public async Task EditMessage(Context ctx)
...
+        var prepend = ctx.MatchFlag("prepend");
+        
+        // Grab the original message content and new message content
+        var originalContent = originalMsg.Content;
         var newContent = ctx.RemainderOrNull().NormalizeLineEndSpacing();
+ 
+        // Append or prepend the new content to the original message content if needed.
+        // If no flag is supplied, the new contents will completly overwrite the old contents
+        // If both flags are specified. the message will be prepended AND appended
+        newContent = append ? $"{originalContent} {newContent}" : $"{newContent}";
```
spiralw, line 69: this should really not be an inline conditional

spiralw (changes requested):
Can you add the new flags to the documentation (`/docs/content/tips-and-tricks.md`)?

PluralKit.Bot/Commands/Message.cs
UsernameNotAvalible: I changed it from an inline conditional to an if-then statement. Also, the documentation is updated now

PR #437, opened 2022-03-31
Add -raw flag to member list

UsernameNotAvalible:
This adds a -raw flag to the Pk;ls command. The -raw flag will list the names (not display names) of all members in the system inside a code block (so that they aren't formatted with markdown). The flags -by-id (shortened to -bid) and -by-display-name (shortened to -bdn) can both be used with the raw flag to add the member IDs or Display names to the output

### DIFF
--- a/PluralKit.Bot/Commands/SystemList.cs
+++ b/PluralKit.Bot/Commands/SystemList.cs
@@ -53,7 +53,7 @@ public class SystemList
         return title.ToString();
     }
 
-    private string GenerateRawList(Context ctx, List members) {
+    private string GenerateRawList(Context ctx, List<ListedMember> members) {
         // Get flags
         var showId = ctx.MatchFlag("show-id", "id");
         var showDispName = ctx.MatchFlag("show-display-name", "show-dn", "dn");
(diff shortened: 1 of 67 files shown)


PluralKit.Bot/Commands/SystemList.cs
```diff
@@ -18,6 +18,23 @@ public async Task MemberList(Context ctx, PKSystem target)
         // - RenderMemberList checks the indivual privacy for each member (NameFor, etc)
         // the own system is always allowed to look up their list
         var opts = ctx.ParseListOptions(ctx.DirectLookupContextFor(target.Id));
+        if (ctx.MatchFlag("raw")) {
+            var members = (await ctx.Database.Execute(conn => conn.QueryMemberList(target.Id, opts.ToQueryOptions())))
+                .ToList();
+            
+            var showId = ctx.MatchFlag("by-id") || ctx.MatchFlag("bid") || ctx.MatchFlag("id");
```
spiralw, line 25: generally, `by-x` flags are *sorting* options, not *display* options; either way this needs to use the options from `ParseListOptions`

also worth noting: you can pass multiple arguments to `Match` or `MatchFlag` instead of calling the method multiple times

spiralw (changes requested):
this is missing privacy settings - please read the comment directly above your code

PluralKit.Bot/Commands/SystemList.cs
UsernameNotAvalible: I made the flag checks use a single call to MatchFlag, and i also added a privacy check that *should* work. Also, I changed the option to "show-id" (shortened to "id") and "show-display-name" (shortened to "show-dn" or "dn")

UsernameNotAvalible:
i fixed it (i think)

UsernameNotAvalible:
commit [3a3ffff](https://github.com/xSke/PluralKit/pull/437/commits/3a3ffff43825267f0199638d3332eb2adaf0559b) fixed the privacy, and commits [d9b4d4f](https://github.com/xSke/PluralKit/pull/437/commits/d9b4d4f186c87ef45dd1ac64ddf7acc3b2d3d5c3) and [b0ce0b8](https://github.com/xSke/PluralKit/pull/437/commits/b0ce0b8dcc943fc3d515c87bd8a2ead7495c69b0) fixed more stuff

spiralw:
no, commit 3a3ffff does not fix the privacy. You need to check against `member.NamePrivacy if the "show displayname" flag is not set. as the comment right above your code states, "[the full member list code] checks the indivual privacy for each member (NameFor, etc)"

UsernameNotAvalible:
I finally fixed it :D

PR #438, opened 2022-03-31
Make proxy tags not cap-sensitive

UsernameNotAvalible:
This change is pretty simple. It makes proxy tags not be caps-sensitive, so that typing `TT:hi` could also proxy as a member with proxy tags `tt:text` or `tT:text`. If 2 members have the same proxy tag with different capitalization and you try to proxy using the proxy tags, the bot will prioritize whatever member has proxy tags with capitalization closest to how you capitalized it in the message to proxy

### DIFF
--- a/Myriad/Rest/Types/Requests/MessageRequest.cs
+++ b/Myriad/Rest/Types/Requests/MessageRequest.cs
@@ -10,4 +10,6 @@ public record MessageRequest
     public AllowedMentions? AllowedMentions { get; set; }
     public Embed[]? Embeds { get; set; }
     public MessageComponent[]? Components { get; set; }
+    public Message.Reference? MessageReference { get; set; }
+
 }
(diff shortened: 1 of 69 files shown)


UsernameNotAvalible:
Someone suggested this, so i made it. It could cause problems if someone has 2 members with the same proxy tags with different capitalization, though.

spiralw:
This needs to be an *option* - and by default set to false - whether per member or globally for the system.

UsernameNotAvalible:
What should i name the config option?

UsernameNotAvalible:
I don't think it's possible to access the config info from this context, so it might be near-impossible for me to make this be controlled by a config setting

spiralw:
I'm not very familiar with that code but you should be able to pass it through in `ProxyTagParser.TryMatch`?

UsernameNotAvalible:
I don't think it is possible to have this controlled by a config value without having to rewrite a huge amount of code

spiralw:
oh right, you'd have to add it to MessageContext in SQL

PR #444, opened 2022-05-07
Auto-export on system deletion

u1f408:
This makes PluralKit automatically export the user's system when they run `pk;system delete`. If the export file can not be sent in DMs to the user, system deletion will not occur, unless the `-no-export` flag is specified on command invocation.

### DIFF
--- a/PluralKit.Bot/Commands/SystemEdit.cs
+++ b/PluralKit.Bot/Commands/SystemEdit.cs
@@ -1,7 +1,13 @@
+using System.Text;
 using System.Text.RegularExpressions;
 
 using Myriad.Builders;
+using Myriad.Rest.Exceptions;
+using Myriad.Rest.Types;
+using Myriad.Rest.Types.Requests;
 using Myriad.Types;
+
+using Newtonsoft.Json;
 
 using PluralKit.Core;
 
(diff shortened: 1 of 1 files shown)


PluralKit.Bot/Commands/SystemEdit.cs
```diff
@@ -523,9 +533,44 @@ public async Task Delete(Context ctx, PKSystem target)
...
+                await ctx.Rest.CreateMessage(dm, new MessageRequest { Content = $"<{msg.Attachments[0].Url}>" });
+
+                // If the original message wasn't posted in DMs, send a public reminder
+                if (ctx.Channel.Type != Channel.ChannelType.Dm)
+                    await ctx.Reply($"{Emojis.Success} System deleted. Check your DMs for a copy of your system's data!");
+            }
+            catch (ForbiddenException)
+            {
+                // If user has DMs closed, tell 'em to open them
+                throw new PKError(
```
xSke, line 566: Can you reword this error? If someone with DMs closed is running `pk;s delete`, don't know about the export, and get this error, it doesn't make sense. Maybe have it tell them it tried to send an export file when deleting, and to use the flag if they just want to do it anyway.

PluralKit.Bot/Commands/SystemEdit.cs
```diff
@@ -523,9 +533,44 @@ public async Task Delete(Context ctx, PKSystem target)
...
+                    $"Could not send a copy of your system's data in DMs. Are your DMs closed?");
+            }
+        }
+
+        // Now that we've sent the export data (or been told not to), we can safely delete the system
 
-        await ctx.Reply($"{Emojis.Success} System deleted.");
```
xSke, line 528: If noexport is on then it won't send a confirm message, maybe put this in an else

xSke:
Maybe also change the "are you sure you wanna delete your system" message to also say it'll try to send an export file?

PR #445, opened 2022-05-07
feat: make cancelled commands reply to trigger

Draconizations:
makes all "operation cancelled" errors and variants thereof reply to it's trigger message. this happens both with prompts being timed out and being cancelled manually.

### DIFF
--- a/PluralKit.Bot/CommandSystem/Context/Context.cs
+++ b/PluralKit.Bot/CommandSystem/Context/Context.cs
@@ -77,7 +77,7 @@ public class Context
     internal readonly IDatabase Database;
     internal readonly ModelRepository Repository;
 
-    public async Task<Message> Reply(string text = null, Embed embed = null, AllowedMentions? mentions = null)
+    public async Task<Message> Reply(string text = null, Embed embed = null, AllowedMentions? mentions = null, bool referenceCommand = false)
     {
         var botPerms = await BotPermissions;
 
@@ -92,6 +92,7 @@ public class Context
         {
             Content = text,
             Embeds = embed != null ? new[] { embed } : null,
+            MessageReference = referenceCommand ? new Message.Reference(Message.GuildId, Message.ChannelId, Message.Id) : null,
             // Default to an empty allowed mentions object instead of null (which means no mentions allowed)
             AllowedMentions = mentions ?? new AllowedMentions()
         });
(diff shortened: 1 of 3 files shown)


PluralKit.Bot/Utils/ContextUtils.cs
```diff
@@ -161,7 +161,7 @@ async Task<int> PromptPageNumber()
                         }
                         catch (TimeoutException)
                         {
-                            await ctx.Reply($"{Emojis.Error} Operation timed out, sorry. Try again, perhaps?");
+                            await ctx.Reply($"{Emojis.Error} Operation timed out, sorry. Try again, perhaps?", null, null, true);
```
spiralw, line 164: This should reference the "what page would you like to go to" message, rather than the original command

xSke, line 164: i'd use `referenceCommand: true` just to avoid the `, null, null, ` - makes it clear which parameter is being set

PluralKit.Bot/CommandSystem/Context/Context.cs
```diff
@@ -121,14 +122,18 @@ public async Task Execute<T>(Command? commandDef, Func<T, Task> handler, bool de
...
+        }
         catch (PKError e)
         {
             await Reply($"{Emojis.Error} {e.Message}");
         }
         catch (TimeoutException)
         {
             // Got a complaint the old error was a bit too patronizing. Hopefully this is better?
-            await Reply($"{Emojis.Error} Operation timed out, sorry. Try again, perhaps?");
+            await Reply($"{Emojis.Error} Operation timed out, sorry. Try again, perhaps?", null, null, true);
```
xSke, line 136: i'd use `referenceCommand: true` just to avoid the `, null, null, ` - makes it clear which parameter is being set

spiralw:
```suggestion
            await Reply($"{Emojis.Error} Operation timed out, sorry. Try again, perhaps?", referenceCommand true);
```

spiralw:
hm, now I'm wondering if it would be worth it to have every sent message reply to the original command. would be more in line with how slash commands behave, since we do eventually want to implement those it'd be nice to have similar behaviour

Draconizations:
> hm, now I'm wondering if it would be worth it to have every sent message reply to the original command. would be more in line with how slash commands behave, since we do eventually want to implement those it'd be nice to have similar behaviour

I was also thinking about that honestly ^^; I can close this PR and PR that instead if that's preferred

spiralw:
sure :+1:

PR #447, opened 2022-05-08
feat: add `pk;reproxy`

u1f408:
Adds `pk;reproxy` (`pk;rp`) to reproxy the last sent message, or a linked message (in the same way `pk;edit` does) using the given member, which can be specified as a member name or ID.

This is probably not the best way to do this, very happy to make any changes.

### DIFF
--- a/PluralKit.Bot/CommandMeta/CommandHelp.cs
+++ b/PluralKit.Bot/CommandMeta/CommandHelp.cs
@@ -86,6 +86,7 @@ public partial class CommandTree
     public static Command Explain = new Command("explain", "explain", "Explains the basics of systems and proxying");
     public static Command Message = new Command("message", "message <id|link> [delete|author]", "Looks up a proxied message");
     public static Command MessageEdit = new Command("edit", "edit [link] <text>", "Edit a previously proxied message");
+    public static Command MessageReproxy = new Command("reproxy", "reproxy [link] <member>", "Reproxy a previously proxied message using a different member");
     public static Command ProxyCheck = new Command("debug proxy", "debug proxy [link|reply]", "Checks why your message has not been proxied");
     public static Command LogChannel = new Command("log channel", "log channel <channel>", "Designates a channel to post proxied messages to");
     public static Command LogChannelClear = new Command("log channel", "log channel -clear", "Clears the currently set log channel");
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -62,8 +62,7 @@ public class ProxiedMessage
             throw new PKError("The message is too old to be reproxied.");
 
         // Get target member ID
-        var target = await ctx.MatchMember();
-        var previousPtr = ctx.Parameters._ptr;
+        var target = await ctx.MatchMember(restrictToSystem: ctx.System.Id);
         if (target == null)
             throw new PKError("Could not find a member to reproxy the message with.");
 
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -182,7 +182,7 @@ public class ProxiedMessage
         }
 
         var msgTimestamp = DiscordUtils.SnowflakeToInstant(msg.Message.Mid);
-        if (SystemClock.Instance.GetCurrentInstant() - msgTimestamp > timeout)
+        if (isReproxy && SystemClock.Instance.GetCurrentInstant() - msgTimestamp > timeout)
             throw new PKError($"The message is too old to be {editTypeAction}.");
 
         return msg;
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -56,6 +56,11 @@ public class ProxiedMessage
         if (ctx.System.Id != msg.System?.Id)
             throw new PKError("Can't reproxy a message sent by a different system.");
 
+        // Bail if the message is more than 2 minutes old
+        var msgTimestamp = Instant.FromUnixTimeMilliseconds((long)(msg.Message.Mid >> 22) + 1420070400000);
+        if (msgTimestamp.Plus(Duration.FromSeconds(60 * 2)) < SystemClock.Instance.GetCurrentInstant())
+            throw new PKError("The message is too old to be reproxied.");
+
         // Get target member ID
         var target = await ctx.MatchMember();
         var previousPtr = ctx.Parameters._ptr;
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -50,7 +50,7 @@ public class ProxiedMessage
 
     public async Task ReproxyMessage(Context ctx)
     {
-        var msg = await GetMessageToEdit(ctx, ReproxyTimeout);
+        var msg = await GetMessageToEdit(ctx, ReproxyTimeout, true);
 
         if (ctx.System.Id != msg.System?.Id)
             throw new PKError("Can't reproxy a message sent by a different system.");
@@ -85,7 +85,7 @@ public class ProxiedMessage
 
     public async Task EditMessage(Context ctx)
     {
-        var msg = await GetMessageToEdit(ctx, EditTimeout);
+        var msg = await GetMessageToEdit(ctx, EditTimeout, false);
 
         if (ctx.System.Id != msg.System?.Id)
             throw new PKError("Can't edit a message sent by a different system.");
@@ -135,8 +135,11 @@ public class ProxiedMessage
         }
     }
 
-    private async Task<FullMessage> GetMessageToEdit(Context ctx, Duration timeout)
+    private async Task<FullMessage> GetMessageToEdit(Context ctx, Duration timeout, bool isReproxy)
     {
+        var editType = isReproxy ? "reproxy" : "edit";
+        var editTypeAction = isReproxy ? "reproxied" : "edited";
+
         // todo: is it correct to get a connection here?
         await using var conn = await ctx.Database.Obtain();
         FullMessage? msg = null;
@@ -152,15 +155,15 @@ public class ProxiedMessage
         if (msg == null)
         {
             if (ctx.Guild == null)
-                throw new PKSyntaxError("You must use a message link to edit messages in DMs.");
+                throw new PKSyntaxError($"You must use a message link to {editType} messages in DMs.");
 
             var recent = await FindRecentMessage(ctx, timeout);
             if (recent == null)
-                throw new PKSyntaxError("Could not find a recent message to edit.");
+                throw new PKSyntaxError($"Could not find a recent message to {editType}.");
 
             msg = await ctx.Repository.GetMessage(conn, recent.Mid);
             if (msg == null)
-                throw new PKSyntaxError("Could not find a recent message to edit.");
+                throw new PKSyntaxError($"Could not find a recent message to {editType}.");
         }
 
         if (msg.Message.Channel != ctx.Channel.Id)
@@ -180,7 +183,7 @@ public class ProxiedMessage
 
         var msgTimestamp = DiscordUtils.SnowflakeToInstant(msg.Message.Mid);
         if (SystemClock.Instance.GetCurrentInstant() - msgTimestamp > timeout)
-            throw new PKError("The message is too old to be edited.");
+            throw new PKError($"The message is too old to be {editTypeAction}.");
 
         return msg;
     }
--- a/PluralKit.Bot/Commands/Message.cs
+++ b/PluralKit.Bot/Commands/Message.cs
@@ -22,6 +22,7 @@ namespace PluralKit.Bot;
 public class ProxiedMessage
 {
     private static readonly Duration EditTimeout = Duration.FromMinutes(10);
+    private static readonly Duration ReproxyTimeout = Duration.FromMinutes(1);
 
     // private readonly IDiscordCache _cache;
     private readonly ModelRepository _repo;
@@ -49,15 +50,10 @@ public class ProxiedMessage
 
     public async Task ReproxyMessage(Context ctx)
     {
-        var msg = await GetMessageToEdit(ctx);
+        var msg = await GetMessageToEdit(ctx, ReproxyTimeout);
 
         if (ctx.System.Id != msg.System?.Id)
             throw new PKError("Can't reproxy a message sent by a different system.");
-
-        // Bail if the message is more than 2 minutes old
-        var msgTimestamp = Instant.FromUnixTimeMilliseconds((long)(msg.Message.Mid >> 22) + 1420070400000);
-        if (msgTimestamp.Plus(Duration.FromSeconds(60 * 2)) < SystemClock.Instance.GetCurrentInstant())
-            throw new PKError("The message is too old to be reproxied.");
 
         // Get target member ID
         var target = await ctx.MatchMember(restrictToSystem: ctx.System.Id);
@@ -89,7 +85,7 @@ public class ProxiedMessage
 
     public async Task EditMessage(Context ctx)
     {
-        var msg = await GetMessageToEdit(ctx);
+        var msg = await GetMessageToEdit(ctx, EditTimeout);
 
         if (ctx.System.Id != msg.System?.Id)
             throw new PKError("Can't edit a message sent by a different system.");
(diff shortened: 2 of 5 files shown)


PluralKit.Bot/CommandMeta/CommandHelp.cs
```diff
@@ -86,6 +86,7 @@ public partial class CommandTree
     public static Command Explain = new Command("explain", "explain", "Explains the basics of systems and proxying");
     public static Command Message = new Command("message", "message <id|link> [delete|author]", "Looks up a proxied message");
     public static Command MessageEdit = new Command("edit", "edit [link] <text>", "Edit a previously proxied message");
+    public static Command MessageReproxy = new Command("reproxy", "reproxy [link] <member>", "Reproxy a previously proxied message using a different member");
```
spiralw, line 89: needs added to docs website

PluralKit.Bot/Commands/Message.cs
```diff
@@ -13,6 +13,8 @@
 
 using NodaTime;
 
+using IMetrics = App.Metrics.IMetrics;
```
spiralw, line 16: I think you can just `using App.Metrics;`?

PluralKit.Bot/Commands/Message.cs
```diff
@@ -21,24 +23,98 @@ public class ProxiedMessage
...
+            throw new PKError("Could not reproxy message.");
+
+        // Get a MessageContext for the original message
+        MessageContext? msgCtx =
+               await _repo.GetMessageContext(msg.Message.Sender, msg.Message.Guild!.Value, originalMsg.ChannelId);
+
+        try
+        {
+            var messageChannel = await _cache.GetChannel(originalMsg.ChannelId);
+            var rootChannel = await _cache.GetRootChannel(originalMsg.ChannelId);
```
spiralw, line 80: you can't use cache here, clusters don't (currently) share cache - see GetMessage

PluralKit.Bot/Commands/Message.cs
```diff
@@ -21,24 +23,98 @@ public class ProxiedMessage
...
         _rest = rest;
         _webhookExecutor = webhookExecutor;
+        _repo = repo;
         _logChannel = logChannel;
-        // _cache = cache;
+        _cache = cache;
+        _metrics = metrics;
+    }
+
+    public async Task ReproxyMessage(Context ctx)
```
spiralw, line 50: make sure this method makes all the necessary permission checks that need to happen when re-proxying a message from DMs / another server. it's possible someone might try to reproxy a message after having send message permissions removed

spiralw (changes requested):
- I would prefer if this code was in ProxyService, a lot of the logic is duplicated / you are missing some important stuff from `HandleProxyExecutedActions`

- should the reproxy command check if the message it's about to reproxy is the last message sent in the channel? it'd be fairly confusing to re-order messages after the fact. maybe a time check (this message was sent x time ago)?

PluralKit.Bot/Commands/Message.cs
u1f408: Doing that makes the compiler yell because `App.Metrics` has an incompatible `IClock` definition (which is pulled in by `using NodaTime;`)

u1f408:
Will move into `ProxyService` - wasn't sure where the best place to put this was!

PluralKit.Bot/Commands/Message.cs
spiralw: in that case I'd change to use `SystemClock.Instance` instead of the DI IClock, but it's probably irrelevant now since the code's getting moved to ProxyService anyway

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,65 @@ public class ProxiedMessage
...
+        if (target == null)
+            throw new PKError("Could not find a member to reproxy the message with.");
+
+        // Fetch members and get the ProxyMember for `target`
+        List <ProxyMember> members;
+        using (_metrics.Measure.Timer.Time(BotMetrics.ProxyMembersQueryTime))
+            members = (await _repo.GetProxyMembers(ctx.Author.Id, msg.Message.Guild!.Value)).ToList();
+        var match = members.Find(x => x.Id == target.Id);
+        if (match == null)
+            throw new PKError("Could not find a member to reproxy the message with.");
```
u1f408, line 71: not sure whether to move this section into the reproxy method in ProxyService, it kinda makes more sense to us to have the member searching where it is currently? (that being said, we're happy to move it if y'all think it makes more sense)

u1f408:
a time check on reproxying is probably a good idea, tbh - i'm hesitant to make it only work if the message was the latest in the channel, as you'd never be able to _use_ the reproxy in faster-moving channels at that point. maybe "can only reproxy if message is less than 1min old" would work?

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -188,6 +188,56 @@ public bool ShouldProxy(Channel channel, Message msg, MessageContext ctx)
...
+            Attachments = originalMsg.Attachments!,
+            FileSizeLimit = guild.FileSizeLimit(),
+            Embeds = originalMsg.Embeds!.ToArray(),
+            Stickers = originalMsg.StickerItems!,
+            AllowEveryone = allowEveryone
+        });
+
+        var sentMessage = await HandleProxyExecutedActions(ctx, autoproxySettings, trigger, proxyMessage, match, deletePrevious: false);
+        await _rest.DeleteMessage(originalMsg.ChannelId!, originalMsg.Id!);
+        await _logChannel.LogMessage(ctx, sentMessage, trigger, proxyMessage, originalMsg.Content!);
```
spiralw, line 238: `HandleProxyExecutedActions` already does all of this, not sure why it's duplicated here

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,70 @@ public class ProxiedMessage
...
+        if (ctx.System.Id != msg.System?.Id)
+            throw new PKError("Can't reproxy a message sent by a different system.");
+
+        // Bail if the message is more than 2 minutes old
+        var msgTimestamp = Instant.FromUnixTimeMilliseconds((long)(msg.Message.Mid >> 22) + 1420070400000);
+        if (msgTimestamp.Plus(Duration.FromSeconds(60 * 2)) < SystemClock.Instance.GetCurrentInstant())
+            throw new PKError("The message is too old to be reproxied.");
+
+        // Get target member ID
+        var target = await ctx.MatchMember();
```
spiralw, line 65: I think most people expect this command to work with proxy tags instead of a member name/id

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,70 @@ public class ProxiedMessage
...
+            throw new PKError("Could not find a member to reproxy the message with.");
+
+        try
+        {
+            await _proxy.ExecuteReproxy(ctx.Message, msg.Message, match);
+
+            if (ctx.Guild == null)
+                await _rest.CreateReaction(ctx.Channel.Id, ctx.Message.Id, new Emoji { Name = Emojis.Success });
+            if ((await ctx.BotPermissions).HasFlag(PermissionSet.ManageMessages))
+                await _rest.DeleteMessage(ctx.Channel.Id, ctx.Message.Id);
```
spiralw, line 85: this is already done by HandleProxyExecutedActions, you just need to delete the original proxied message

PluralKit.Bot/Commands/Message.cs
spiralw: it's probably fine here, yeah

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,70 @@ public class ProxiedMessage
...
+        // Get target member ID
+        var target = await ctx.MatchMember();
+        var previousPtr = ctx.Parameters._ptr;
+        if (target == null)
+            throw new PKError("Could not find a member to reproxy the message with.");
+
+        // Fetch members and get the ProxyMember for `target`
+        List <ProxyMember> members;
+        using (_metrics.Measure.Timer.Time(BotMetrics.ProxyMembersQueryTime))
+            members = (await _repo.GetProxyMembers(ctx.Author.Id, msg.Message.Guild!.Value)).ToList();
```
spiralw, line 73: this is fairly expensive, is there a reason why you can't directly use the PKMember from MatchMember?

PluralKit.Bot/Commands/Message.cs
spiralw, line 65:
```suggestion
        var target = await ctx.MatchMember(restrictToSystem: ctx.System.Id);
```

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,70 @@ public class ProxiedMessage
...
+            throw new PKError("Can't reproxy a message sent by a different system.");
+
+        // Bail if the message is more than 2 minutes old
+        var msgTimestamp = Instant.FromUnixTimeMilliseconds((long)(msg.Message.Mid >> 22) + 1420070400000);
+        if (msgTimestamp.Plus(Duration.FromSeconds(60 * 2)) < SystemClock.Instance.GetCurrentInstant())
+            throw new PKError("The message is too old to be reproxied.");
+
+        // Get target member ID
+        var target = await ctx.MatchMember();
+        var previousPtr = ctx.Parameters._ptr;
```
spiralw, line 66:
```suggestion
```

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -188,6 +188,56 @@ public bool ShouldProxy(Channel channel, Message msg, MessageContext ctx)
...
+    {
+        var originalMsg = await _rest.GetMessageOrNull(msg.Channel, msg.Mid);
+        if (originalMsg == null)
+            throw new PKError("Could not reproxy message.");
+
+        // Get a MessageContext for the original message
+        MessageContext ctx =
+            await _repo.GetMessageContext(msg.Sender, msg.Guild!.Value, msg.Channel);
+
+        var autoproxySettings = await _repo.GetAutoproxySettings(ctx.SystemId.Value, msg.Guild!.Value, null);
```
spiralw, line 201: this seems unused?

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -188,6 +188,56 @@ public bool ShouldProxy(Channel channel, Message msg, MessageContext ctx)
...
     }
 
+    public async Task ExecuteReproxy(Message trigger, PKMessage msg, ProxyMember member)
+    {
+        var originalMsg = await _rest.GetMessageOrNull(msg.Channel, msg.Mid);
+        if (originalMsg == null)
+            throw new PKError("Could not reproxy message.");
+
+        // Get a MessageContext for the original message
+        MessageContext ctx =
```
spiralw, line 198: should check `ctx.InBlacklist` and throw early

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -338,6 +389,9 @@ Task SaveMessageInDatabase()
 
         async Task DeleteProxyTriggerMessage()
         {
+            if (!deletePrevious)
```
spiralw, line 392: what is this for?

spiralw:
> maybe "can only reproxy if message is less than 1min old" would work?

sure, that's fine

PluralKit.Bot/Proxy/ProxyService.cs
u1f408: the task in HandleProxyExecuteActions will delete the reproxied message if we delete the trigger, hence some of the duplication of message deletion stuff & this check to return early in that task

PluralKit.Bot/Proxy/ProxyService.cs
u1f408: autoproxySettings is only here because it needs to be passed to HandleProxyExecutedActions, we don't actually need it otherwise

PluralKit.Bot/Commands/Message.cs
u1f408: we get a ProxyMember here because that has the utility methods for retrieving server name/avatar/etc - can figure out getting those another way if this is too expensive?

PluralKit.Bot/Commands/Message.cs
u1f408: problem with using proxy tags here is that we then have no easy way to distinguish between a prefix and a suffix. if proxy tags were prefix only (or suffix only)  then that would work, but they're not - so you'd have to parse them like the proxy tag set method does (with the `text` sentinel). looking up a member by name/ID makes this a lot easier both in implementation and in usage

PluralKit.Bot/Commands/Message.cs
u1f408: this is done here so that this commands acts the same way as `pk;edit` - deleting the trigger if in a channel, marking it as successful with an emoji in DMs. figured it's best to have consistency between the message-editing-related commands?

(that is also why we added the "don't delete the trigger" override flag in HandleProxyExecutedActions)

spiralw: ohh, okay, that makes sense

PluralKit.Bot/Commands/Message.cs
spiralw: I looked at the code and it's probably fine like this

PluralKit.Bot/Commands/Message.cs
spiralw: I was going to suggest using a ProxyTagParser then I remembered this method doesn't actually take any content. makes sense to use a member name/id then

PluralKit.Bot/Commands/Message.cs
u1f408: taking a quick look at this, removing the DI IClock (and the `NodaTime` import) would require rewriting unrelated stuff around the `pk;edit` timeout, and we don't really wanna mess with that in this PR tbh!

spiralw: oh, that's fine. I'll fix it myself when I merge the PR then

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,69 @@ public class ProxiedMessage
...
+
+    public async Task ReproxyMessage(Context ctx)
+    {
+        var msg = await GetMessageToEdit(ctx);
+
+        if (ctx.System.Id != msg.System?.Id)
+            throw new PKError("Can't reproxy a message sent by a different system.");
+
+        // Bail if the message is more than 2 minutes old
+        var msgTimestamp = Instant.FromUnixTimeMilliseconds((long)(msg.Message.Mid >> 22) + 1420070400000);
```
spiralw, line 58: is it possible to move this check inside of `GetMessageToEdit`? maybe add another line next to EditTimeout and pass it into the method

PluralKit.Bot/Commands/Message.cs
```diff
@@ -22,23 +24,69 @@ public class ProxiedMessage
...
+        _repo = repo;
         _logChannel = logChannel;
         // _cache = cache;
+        _metrics = metrics;
+        _proxy =  proxy;
+    }
+
+    public async Task ReproxyMessage(Context ctx)
+    {
+        var msg = await GetMessageToEdit(ctx);
```
spiralw, line 52: the error messages in this method all mention editing rather than reproxying

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -307,8 +361,9 @@ private string FixSameNameInner(string name)
     public static bool IsUnlatch(Message message)
         => message.Content.StartsWith(@"\\") || message.Content.StartsWith("\\\u200b\\");
 
-    private async Task HandleProxyExecutedActions(MessageContext ctx, AutoproxySettings autoproxySettings,
-                                                Message triggerMessage, Message proxyMessage, ProxyMatch match)
+    private async Task<PKMessage> HandleProxyExecutedActions(MessageContext ctx, AutoproxySettings autoproxySettings,
```
spiralw, line 364: since `sentMessage` isn't being used upstream, can you revert this change?

spiralw (changes requested):

PluralKit.Bot/Commands/Message.cs
u1f408: not sure what's best to do with the errors in `GetMessageToEdit` - passing in a `bool isReproxy` and changing the messages based on that? 🤷

spiralw: yeah, that should be fine

PR #451, opened 2022-05-22
Added pinging when replypinging

Tiefseetauchner:
I had the issue, that I got modpinged by a systemmember, and it essentially turned into a ghostping and I didn't notice that I got modpinged

This'll add a ping at the beginning of the message, when the ping reply is enabled and the person isn't being pinged in the message

Something like this:

image

Here I added Replypinging to the message, but I felt like that wasn't necessary. 

I don't know if it's really needed or wanted, but I'm putting it out there, mainly cuz I actually wanted something to program again :)

Also sry for those two first commits, it was an old fork, I didn't feel like making a new one, so I just merged your upstream :/

Best, Lena!


xSke:
instead of checking for a @ in the content, can we check whether it actually pinged? including when someone replies with the ping enabled. 

i ask because i don't remember if that information is actually exposed in the event.

Tiefseetauchner:
Indeed. The pinging reply puts the mention in [0] of Mentions, but it'd be the same if I replied  without a ping to someone with the message

Hi @LukeChriswalker

So in that case, the result would be

*@LukeChriswalker*

Hi @LukeChriswalker

And that wouldn't be intended behaviour for me. Edge case but imho worth implementing, so the bot doesn't feel trashy

Tiefseetauchner:
Any update on this? Do you need something, can I advance this? I got ghost pinged again by a system member, that's why I'm "reopening" it

I'll fetch the upstream and look at merge conflicts for now, just ping me if there's anything I can do that's needed to merge this here

kloenk (approved):
From what I can tell the code looks reasonable, and interaction with systems, not beeing ghost pinged would be a great improvement

PluralKit.Bot/Proxy/ProxyService.cs
```diff
@@ -148,11 +148,21 @@ public bool ShouldProxy(Channel channel, Message msg, MessageContext ctx)
...
         {
             var repliedTo = trigger.ReferencedMessage.Value;
             if (repliedTo != null)
             {
+                if (trigger.Mentions.Length > 0 
+                    && repliedTo.Author.Id == trigger.Mentions[0].Id
+                    && !(trigger.Content.Contains($"<@{repliedTo.Author.Id}>")
+                         || trigger.Content.Contains($"<@!{repliedTo.Author.Id}>")))
+                {
+                    content = $"*<@{repliedTo.Author.Id}>*\n";
```
thetayloredman, line 163:
```suggestion
                    content = $"<@{repliedTo.Author.Id}>\n";
```

Italics around an @-mention do nothing afaik, and it wastes 2 characters

Tiefseetauchner: It does on mobile



I caj't remove it easily on mobile but I can get to it when I'm at home, if it should be removed

thetayloredman: I can't see it being needed either way

Tiefseetauchner: I don't see a reason to remove it, can you explain why you would?

PR #462, opened 2022-06-11
fix(docs): slightly clarify system command targeting

u1f408:
Discussed a bit in `#bug-reports-and-errors` in the support server, clarifies that entering an ID for `pk;system` commands is optional when targeting your own system

### DIFF
--- a/docs/content/command-list.md
+++ b/docs/content/command-list.md
@@ -32,7 +32,7 @@ Some arguments indicate the use of specific Discord features. These include:
 
 # Commands
 ## System commands
-*Optionally replace `[system]` with a @mention, Discord account ID, or 5-character ID. For most commands, adding `-clear` will clear/delete the field.*
+*To target your own PluralKit system with these commands, you do not need to enter anything for `[system]`. To target another system, replace `[system]` with that systems' 5-character ID, a Discord account ID, or an @mention. For most commands, adding `-clear` will clear/delete the field.*
 - `pk;system [system]` - Shows information about a system.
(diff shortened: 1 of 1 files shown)


macks2008:
Here's what threw me (the uninitiated newb that prompted this issue) off originally, in case you want to work it in somehow: I assumed the name of the system could be used to target it. I'm getting the impression, based on the conversation on discord, that's wrong and that the ID is the way to go normally.
(why would you even want to target someone else's system though, if I may ask? again,complete newb here)

macks2008:
@spiralw How about "system name cannot be used to target a system"?
Come to think of it, would "display name" be more accurate, if there is nowhere that the name is used for targeting?

macks2008:
@spiralw Yeah... How many guilds are you in now? I think I saw a six digit number somewhere about that, but I can't find it now
[Although on the other hand...](https://xkcd.com/1172/)

u1f408:
Updated the phrasing a bit, how's that?

pulchra-mentis:
Should `systems'` actually be `system's` since it is in reference to a single PK system?

u1f408:
> Should `systems'` actually be `system's` since it is in reference to a single PK system?

Ack! Yes, it should, I typed too fast and didn't notice /lh

pulchra-mentis:
> > Should `systems'` actually be `system's` since it is in reference to a single PK system?
> 
> Ack! Yes, it should, I typed too fast and didn't notice /lh

All good! We were an English research assistant back in our undergraduate program, so we tend to catch these things.

Thanks for the work on this! It looks really good!

PR #470, opened 2022-07-08
Implement media resizing using Discord media cache

pulchra-mentis:
Improve performance by offloading image scaling to Discord
Only apply the resizing to images stored in the Discord CDN
Set the format of the resized images to WebP for improved performance

Backported from [currently fronting](https://github.com/systemic-chaos/currently-fronting)


dashboard/src/api/resize-media.ts
```diff
@@ -0,0 +1,18 @@
+const discordCDNAttachmentRegex =
+  /^https:\/\/cdn\.discordapp\.com\/attachments\/(.+)$/i
```
u1f408, line 2: IMO this regex should end in `([^?]+)` (w/ no `$` line-end assertion) as to not capture any existing query parameters - even though there wouldn't usually be a query string, this would totally break if there was one

pulchra-mentis: Good catch! We'll take care of it and push sometime this weekend. (Been feeling a bit under the weather thus far today, although it's likely temporary.)

We're in the process of implementing Jest unit tests in our `currently fronting` project, so we'll add one to handle that scenario.

@Draconizations Any need/want/desire to have some Jest tests added to this repo as well? (The test cases in this instance should be identical across repos.) We can add the associated code for GitHub Actions as well, if desired.

pulchra-mentis: Implemented! Thanks again! (Apparently "this weekend" took longer to arrive than expected. We appreciate the patience.)

Draconizations:
Looks good! I'll merge this now. And no worries about the wait time, real life just gets in the way sometimes.

PR #480, opened 2022-08-27
Some small features

Draconizations:
Three main things have been added here:
- added displaying the amount of characters a string field takes up. both when setting it, and when pulling up that field's information.
- allow `pk;random` for other systems. (using `pk;s system random`). because a `pk;s` variation of the command has been added, documentation has changed to show the `pk;s random` variant over `pk;random`. but `pk;random` will still work as a command.
- added a command for systems, members and groups that just prints the raw id of that item, no formatting.

I tried to include docs as thoroughly as possible but it's possible I might've missed something. I'm also not sure about the id commands *only* printing the id, nothing else, as it might be confusing. But I left it as is in case it's fine.

### DIFF
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -159,7 +159,8 @@ public class Groups
                         + $"To clear it, type `pk;group {reference} displayname -clear`."
                         + $"To print the raw display name, type `pk;group {reference} displayname -raw`.");
 
-                eb.Footer(new Embed.EmbedFooter($"Using {target.DisplayName.Length}/{Limits.MaxGroupNameLength} characters."));
+                if (ctx.System?.Id == target.System)
+                    eb.Footer(new Embed.EmbedFooter($"Using {target.DisplayName.Length}/{Limits.MaxGroupNameLength} characters."));
 
                 await ctx.Reply(embed: eb.Build());
             }
@@ -579,7 +580,7 @@ public class Groups
 
     public async Task DisplayId(Context ctx, PKGroup target)
     {
-        await ctx.Reply($"{target.Hid}");
+        await ctx.Reply(target.Hid);
     }
 
     private async Task<PKSystem> GetGroupSystem(Context ctx, PKGroup target)
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -577,6 +577,11 @@ public class Groups
         await ctx.Reply($"{Emojis.Success} Group deleted.");
     }
 
+    public async Task DisplayId(Context ctx, PKGroup target)
+    {
+        await ctx.Reply($"{target.Hid}");
+    }
+
     private async Task<PKSystem> GetGroupSystem(Context ctx, PKGroup target)
     {
         var system = ctx.System;
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -118,7 +118,7 @@ public class Groups
 
         await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { Name = newName });
 
-        await ctx.Reply($"{Emojis.Success} Group name changed from **{target.Name}** to **{newName}**.");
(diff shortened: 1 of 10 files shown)


PluralKit.Bot/Commands/Groups.cs
```diff
@@ -217,7 +219,8 @@ await ctx.Reply(embed: new EmbedBuilder()
                         $"To print the description with formatting, type `pk;group {target.Reference(ctx)} description -raw`."
                         + (ctx.System?.Id == target.System
                             ? $" To clear it, type `pk;group {target.Reference(ctx)} description -clear`."
-                            : "")))
+                            : "")
+                            + $" Using {target.Description.Length}/{Limits.MaxDescriptionLength} characters."))
```
spiralw, line 223: this should probably be own-system only

Draconizations: Gotcha, I applied the same pattern to members/systems too so I’ll change that as well

PluralKit.Bot/Commands/MemberEdit.cs
```diff
@@ -333,7 +335,9 @@ private async Task<EmbedBuilder> CreateMemberNameInfoEmbed(Context ctx, PKMember
         var eb = new EmbedBuilder()
             .Title("Member names")
             .Footer(new Embed.EmbedFooter(
-                $"Member ID: {target.Hid} | Active name in bold. Server name overrides display name, which overrides base name."));
+                $"Member ID: {target.Hid} | Active name in bold. Server name overrides display name, which overrides base name."
+                + (target.DisplayName != null ? $" Using {target.DisplayName.Length}/{Limits.MaxMemberNameLength} characters for the display name." : "")
```
spiralw, line 339: this should be own-system only, leaks the existence of a private name

PluralKit.Bot/Commands/Random.cs
```diff
@@ -15,60 +15,79 @@ public Random(EmbedService embeds)
...
+        var members = await ctx.Repository.GetSystemMembers(target.Id).ToListAsync();
 
         if (!ctx.MatchFlag("all", "a"))
             members = members.Where(m => m.MemberVisibility == PrivacyLevel.Public).ToList();
+        else
+            ctx.CheckOwnSystem(target);
 
         if (members == null || !members.Any())
             throw new PKError(
                 "Your system has no members! Please create at least one member before using this command.");
```
spiralw, line 34: this error needs to be updated to specify someone else's system

PluralKit.Bot/Commands/Random.cs
```diff
@@ -15,60 +15,79 @@ public Random(EmbedService embeds)
...
-        var groups = await ctx.Repository.GetSystemGroups(ctx.System.Id).ToListAsync();
+        var groups = await ctx.Repository.GetSystemGroups(target.Id).ToListAsync();
         if (!ctx.MatchFlag("all", "a"))
             groups = groups.Where(g => g.Visibility == PrivacyLevel.Public).ToList();
+        else
+            ctx.CheckOwnSystem(target);
 
         if (groups == null || !groups.Any())
             throw new PKError(
                 "Your system has no groups! Please create at least one group before using this command.");
```
spiralw, line 56: same as above

PluralKit.Bot/Commands/Groups.cs
```diff
@@ -577,6 +577,11 @@ public async Task DeleteGroup(Context ctx, PKGroup target)
         await ctx.Reply($"{Emojis.Success} Group deleted.");
     }
 
+    public async Task DisplayId(Context ctx, PKGroup target)
+    {
+        await ctx.Reply($"{target.Hid}");
```
spiralw, line 582: it doesn't really make sense to use string formatting if you're just going to put a single string in it
either way, this probably should be in a codeblock, similar to other plaintext output

u1f408: These are specifically not in a codeblock because the mobile clients will copy the backticks - which partially defeats the purpose of having these commands (to make it easier to copy IDs on mobile)

spiralw: oh :/

PR #483, opened 2022-09-05
fix(bot): Proxy tag size limit

Oliver-makes-code:
Having a proxy tag that's too big causes the bot to throw an error, I went in and created a size limit of 64 for prefix and suffix separately, as that sounded reasonable.

### DIFF
--- a/PluralKit.Bot/Commands/MemberProxy.cs
+++ b/PluralKit.Bot/Commands/MemberProxy.cs
@@ -17,6 +17,8 @@ public class MemberProxy
             if (prefixAndSuffix.Length == 1) prefixAndSuffix = prefixAndSuffix[0].Split("TEXT");
             if (prefixAndSuffix.Length < 2) throw Errors.ProxyMustHaveText;
             if (prefixAndSuffix.Length > 2) throw Errors.ProxyMultipleText;
+            if (prefixAndSuffix[0].length > 64) throw Errors.ProxyTooBig;
+            if (prefixAndSuffix[1].length > 64) throw Errors.ProxyTooBig;
             return new ProxyTag(prefixAndSuffix[0], prefixAndSuffix[1]);
         }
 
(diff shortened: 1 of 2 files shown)


Oliver-makes-code:
If a different size is needed, then feel free to tell me

spiralw:
> If a different size is needed, then feel free to tell me

The API has a limit of 100 characters for the entire tag including `text`. I think it's in Limits class, or it should be if it currently isn't

PluralKit.Bot/Commands/MemberProxy.cs
```diff
@@ -17,6 +17,8 @@ ProxyTag ParseProxyTags(string exampleProxy)
             if (prefixAndSuffix.Length == 1) prefixAndSuffix = prefixAndSuffix[0].Split("TEXT");
             if (prefixAndSuffix.Length < 2) throw Errors.ProxyMustHaveText;
             if (prefixAndSuffix.Length > 2) throw Errors.ProxyMultipleText;
+            if (prefixAndSuffix[0].length > 64) throw Errors.ProxyTooBig;
+            if (prefixAndSuffix[1].length > 64) throw Errors.ProxyTooBig;
```
u1f408, line 21: Please change this to use the same check (and return the same error) as the proxy tag length check [later in this file (in the "add" subcommand)](https://github.com/PluralKit/PluralKit/blob/528da39e66447e0fc8eab2e993516c12a6585bf6/PluralKit.Bot/Commands/MemberProxy.cs#L70-L72)

u1f408 (changes requested):

spiralw:
I just noticed this doesn't build, can you fix the build issues? (and preferably test your code before you PR :p)

spiralw:
fixed in 2745ee61564ef0694d5bae29eea771eaa982d3cd

PR #489, opened 2022-09-27
Fix missing newlines in display name commands' help/preview

KJP12:
Includes...
- `pk;group aaaaa displayname`
- `pk;member aaaaa displayname`


spiralw:
thanks!

PR #497, opened 2022-11-12
Rewrite messages storage infrastructure

spiralw:
As we head towards **2 billion** messages worth of history in the production database, it is time to rethink how we handle storage. The current state of the `messages` table in prod has already diverged from the original implementation because of issues at scale; some of these changes are visible in this repo (sorting members by last message was disabled), and some are not (removing foreign key triggers). It has also caused issues with the rest of the database; all user data can theoretically fit in memory for very quick access, but in practice, the majority of the database's cache is used up by messages.

I propose a new design for messages storage that will both solve the above issues and allow for scaling up even higher in the future, if that is ever needed:
- message data will be stored in a separate postgres instance than user data
- messages will be only accessed by `id` key, as to not rely on relational features that may not exist in the *next thing*
- this database will be backed up incrementally
  - since historical data is static (old values may be deleted but never modified), there is no need to back up the entire database multiple times; instead, we can back up changes (added/deleted rows), most likely in the form of WALs
  - it's probably a good idea to start from a base backup again every once in a while. todo: look into how this can be done with the least amount of downtime (postgres docs say the database needs to be shut down while a base backup is being taken)

todo: is it worth partitioning the table? <https://www.postgresql.org/docs/current/ddl-partitioning.html>

There are a few bot features that rely on the current state of the database. Ideally, there should be as little user-facing changes as possible, but there might be some points that can't reasonably be kept.

**It's important to keep the changes to self-host as small as possible.** redis will almost certainly be required for a fully functioning bot, but features that don't require redis should still be functional if it's disabled. This isn't a strict requirement, since it's likely that redis will eventually be required for core bot functionality, but it's worth keeping in mind.

Timeline for database changes:
- [ ] remove database calls that join on `messages` table
- [ ] remove database calls that rely on (original_mid) column index
- [ ] remove database calls that rely on (guild, sender) column index
- [ ] remove database calls that rely on (guild, sender, mid) column index
- [ ] use separate database connection pool to run queries on `messages` table
  - This will require some tweaking to connection count. Right now, each cluster has 50 database connections for everything; we'll have to see how many are required for just messages and for just user data.
- [ ] move `messages` table to its own database
- [ ] set up WAL backups on messages database

We can also take the opportunity here to re-add the previously removed last message sorting. Here, have a checkbox for it:
- [ ] add `last_message_timestamp` to members table

Code changes that will need to be made: 
- [ ] querying FullMessage uses a join against messages table
  - this can be moved to separately querying the message info and the member and system, but after looking into it, most places where this is used actually *don't* need anything more than the raw message
- [ ] looking up messages by original_mid
  - set a redis key to point from original message id to proxied message id, with a reasonable TTL (30 minutes, or so)
  - unsure if there's use cases for looking up a proxied message by original id afterwards
- [ ] logger clean service looks up messages by (user, guild) and (message_id > @ApproxID)
  - set a void redis key for (user, guild) that expires based on how the approximate id is currently calculated
- [ ] edit/reproxy commands look up the latest message by (channel, user)
  - set a redis key mapping (channel, user) to message id, with a TTL based on edit/reproxy timeouts
- [ ] write separate code in `Database.cs` for running queries on messages database

todo: is there anything we're missing here?

The mentioned changes will have to be tested in beta and deployed to production one by one. The PR branch will serve as a "holding bag" of commits to test in beta, and commits will be individually moved to `main` as they go through beta/production deploys.


spiralw:
this is mostly done! there's [some issues](https://github.com/timescale/timescaledb/issues/5075) with timescaledb that we need to figure out, but the only thing left to do is to set up WAL streaming for backups.

spiralw:
this is done now :tada:

PR #503, opened 2022-12-01
fix(dash): Make favicon actually point to a file

rladenson:


rladenson:
The failed format check is for files I didn't modify in this commit

Issue #482, opened 2022-09-02
Slash commands for setup


Madis0:
What if `pk;` was used for message proxying, but slash commands for everything else? 

It would have a lot of benefits, including autocomplete (with parameters), server restrictions (e.g. allow system setup only in a designated channel, disable some non-essential commands entirely), [modal popups](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-modal)...

Related but not a duplicate of #443

u1f408:
There are plans to add slash commands (and message interaction commands) to PluralKit, because they would certainly be useful - however, they're fairly low priority for now (and when we do implement them, they will be in addition to the existing text-based commands, rather than outright replacing them).

PR #513, opened 2023-01-08
Add support for application message commands

### LINKED ISSUE #479: Add context menu interaction equivalent to ❓
Discord supports [message commands](https://discord.com/developers/docs/interactions/application-commands#message-commands) which would make sense as an alternative means to query user info from a message. The UI is a bit better, allowing pluralkit to send a response in the same channel that's visible only to you, and also not letting anyone else see the reaction pop up and disappear. Would be good to have both.

u1f408:
This adds "message commands" (the ones that show up in the context menu under Apps) to do the same things as the existing reaction commands. 

firefox_2Iv7eoHXhi

<details><summary>More screenshots</summary>

firefox_QUkgUmcb73
devenv_b4YA4aa8GQ

</details>

-----

This would close #479, and lays the foundation for implementing #482 if we end up wanting to do that.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -50,3 +50,8 @@ indent_size = 2
 indent_style = space
 indent_size = 4
 tab_width = 4
+
+[*.py]
+indent_style = space
+indent_size = 4
+tab_width = 4
--- a/Myriad/Rest/DiscordApiClient.cs
+++ b/Myriad/Rest/DiscordApiClient.cs
@@ -90,6 +90,11 @@ public class DiscordApiClient
         _client.Delete($"/channels/{channelId}/messages/{messageId}/reactions/{EncodeEmoji(emoji)}",
             ("DeleteAllReactionsForEmoji", channelId));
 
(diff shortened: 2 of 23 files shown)


Kloenk:
(Not yet looked at the C# discord apis, just used the elixir apis until know and just looking into this project)

What I did in my elixir bots is having a bootup function to register the discord / commands. This created a form of checksum of the commands to register, and then compared it to a value stored in my database, So that not every startup requires a resubmit to discord, but I can just change the code, and do not have to run a seperate script, like the python one in this case. maybe that is an option?

Andre601:
Haven't seen if that is doable on webhook messages, but discord allows modal actions (Or whatever they are called).
Like when you right-click a message that it shows a "Apps" option where the bots/apps can add their own stuff. That could be useful in PK for the message lookup, mention, etc.

u1f408:
> Like when you right-click a message that it shows a "Apps" option where the bots/apps can add their own stuff.

That's exactly what this PR is adding :)

PR #514, opened 2023-01-19
feat(bot): change message count sorting

rladenson:
It was pointed out [in the suggestions channel](https://discord.com/channels/466707357099884544/468821582794588160/1065730322379051028) that when sorting with -bmc -r putting members that have 0 messages at the end of the list (where null values go) is uninuitive and it would make more sense to have it at the beginning of the list where it fits in the numerical order instead of being displaced to the end. 
I don't know if the developers agree with this change but I figured since it was an easy PR it wouldn't hurt.


u1f408 (approved):

u1f408:
Merged as 67376be12326b4de0adc636188938d39534af5bd! :)

PR #530, opened 2023-03-12
fix(dash): Group delete prompts

vmorrisonwood:
This pull request corrects the header of the 'Delete group' dialogue on the Dashboard from erroneously saying 'Delete member' instead of 'group' as it should.


spiralw (approved):
thanks!

PR #543, opened 2023-03-24
Several improvements to guide documentation

saizai:
I've:
* fixed typos and display errors (e.g. code blocks and "e.g.")
* simplified or clarified various wording
* added better examples of emoji based tags, and spelled out emoji
* made the examples use a consistent set of names, pronouns, etc. (including respecting the example name changes, display names, keepproxy, etc) and fixed conflicting or inconsistent examples (e.g. John & Craig both having `{text}`)
* fixed incorrect `pk;system timezone` to `pk;config timezone`
* added documentation and examples for 
  * unlatch
  * edit
  * reproxy
  * case-insensitive proxy tags
  * non-`:` prefix, postfix, emoji, & non-bracket-like proxy tags
  * reply-based message query
  * all the shorthand commands except `pk;cfg`
  * `pk; command` being okay
* added warning notices about limitations of some commands, e.g. edit & reproxy timeouts, based on [code](https://github.com/PluralKit/PluralKit/blob/631c6027aba61f85c8f43f4abb8416cc14d1ccb0/PluralKit.Bot/Commands/Message.cs#L211) and some basic testing
* added warning notice about how `pk;m proxy` will require verification if it's replacing rather than adding
* added fully worked examples of this in use, including both usage and PluralKit responses
* added covering sections or moved special cases to subsections

### DIFF
--- a/docs/content/command-list.md
+++ b/docs/content/command-list.md
@@ -31,6 +31,8 @@ Some arguments indicate the use of specific Discord features. These include:
 :::
 
 # Commands
+*Space is allowed after `pk;`; e.g. `pk;system` and `pk; system` have the same effect.
+
 ## System commands
 *To target a specific system, replace `[system]` with that system's 5-character ID, a Discord account ID, or a @mention - note that system names can not be used here. If no system ID is specified, defaults to targeting your own system. For most commands, adding `-clear` will clear/delete the field.*
 *`pk;s` is shorthand for `pk;system`*
@@ -59,7 +61,7 @@ Some arguments indicate the use of specific Discord features. These include:
 
 ## Member commands
 *Replace `<member>` with a member's name, 5-character ID or display name. For most commands, adding `-clear` will clear/delete the field.*
-*`pk;b` is shorthand for `pk;member`*
+*`pk;m` is shorthand for `pk;member`*
 - `pk;member <member>` - Shows information about a member.
 - `pk;member new <name>` - Creates a new system member.
 - `pk;member <member> rename <new name>` - Changes the name of a member.
--- a/docs/content/command-list.md
+++ b/docs/content/command-list.md
@@ -120,7 +120,7 @@ Some arguments indicate the use of specific Discord features. These include:
 - `pk;switch delete all` - Deletes all logged switches.
 
 ## Autoproxy commands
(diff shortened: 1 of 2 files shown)


spiralw:
thanks for the PR! will do a proper review later, but:
- the user guide should always use full commands. shorthands are documented on the "tips and tricks" page
- do not add unrelated documentation to specific commands (member colour is not the right place to explain targeting members by display name, for example)
- do not refer to the bot by its Discord mention (`@PluralKit`) but by name (PluralKit)
- please use the markdown warning format instead of putting a warning emoji in plaintext
- i feel like the autoproxy "example usage" section might be too long and not obvious where one message starts and another one ends, might be worth doing it inline with message/reply?

otherwise looks good :+1:

saizai:
> * the user guide should always use full commands. shorthands are documented on the "tips and tricks" page

I disagree on this.

I agree it should always use full commands as the _first_ examples, and the canonical examples, and I've done that throughout.

However, it should also show what the shorthands are; demonstration by example is clearer than a mention on another page (which might not get read or noticed), and shorthand is part of how to use commands more easily.

> * do not add unrelated documentation to specific commands (member colour is not the right place to explain targeting members by display name, for example)

Agreed and fixed. Not sure how that got in; I intended to put that elsewhere. Have moved it to the top of "member management".

> * do not refer to the bot by its Discord mention (`@PluralKit`) but by name (PluralKit)

Fixed.

> * please use the markdown warning format instead of putting a warning emoji in plaintext

Wasn't aware that was a thing. Fixed to `:warning:`. Ditto for `:information_source:`.

> * i feel like the autoproxy "example usage" section might be too long and not obvious where one message starts and another one ends, might be worth doing it inline with message/reply?

I considered doing that in a few places, but couldn't think of a reasonable (and shorter) way to show either multiple people or replies.

Each message is one line, and it starts with "@Craig#5432 can type this", which I figured should be fairly clear. It's meant to show surround tag, prefix tag, latch, and unlatch functionality; that doesn't need to be as a conversation or replies.

> otherwise looks good 👍

Thanks. 😸

spiralw:
Thanks for your PR. I merged some of your changes in 77142e244fcff2a59defc13333cf1a2638585ce3. We're going to write a contributing guide soon on what changes are or aren't welcome in PRs.

PR #545, opened 2023-04-11
Add LICENSE.md with AGPLv3 text.

MetalJulia:
Just so it's easier to see (as opposed to at the footer of the README) and can be automatically detected. License file pulled from the Markdown link provided on [the GNU website.](https://www.gnu.org/licenses/agpl-3.0.en.html)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,660 @@
+### GNU AFFERO GENERAL PUBLIC LICENSE
+
+Version 3, 19 November 2007
+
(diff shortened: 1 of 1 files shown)


spiralw:
the file should only contain the license text starting from the heading "terms and conditions" and ending before the heading "how to apply these terms to your new program".

MetalJulia:
The heading explicitly states that the license is to be distributed verbatim in lines 8 and 9.

spiralw:
in that case, i don't want to have the license text in the repository

PR #554, opened 2023-05-29
feat: System Guild Icon and System Guild Name

rladenson:
System Guild Icon and System Guild Name are both almost fully implemented, I do want to add System Name Privacy and System Avatar Privacy in this PR as well, which is part of why this is a draft. 

Most places I told it to use system guild name should probably be looked over to check if the behavior I decided on is in fact what's wanted, since this does not fully imitate the member guild name behavior. Member guild names are used for proxying, which means it makes sense to have it more clear on member cards "this is a guild name, this is the normal name" however since System Name is not used in proxying it made the most sense to me to have System Guild Name almost completely cover up System Name in servers that have one. Feedback on this appreciated, feel free to ping me in the server too for more extended discussion. 

As imitating Member Guild Name System Guild Name does not affect pk;msg.

### DIFF
--- a/PluralKit.API/Controllers/v2/SystemControllerV2.cs
+++ b/PluralKit.API/Controllers/v2/SystemControllerV2.cs
@@ -27,7 +27,7 @@ public class SystemControllerV2: PKControllerBase
         if (system == null)
             throw Errors.SystemNotFound;
 
-        return Ok(APIJsonExt.EmbedJson(system.Name ?? $"System with ID `{system.Hid}`", "System"));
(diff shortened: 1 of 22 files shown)


rladenson:
I just realized this won't run properly if you try to test it bc i updated the targetschemaversion to include the privacy migration, the file of which is not committed yet.... sorry :p

rladenson (commented):
Found bug and forgot to update docs, don't merge yet

rladenson:
Ok those should both be fixed and it should be good to merge now

u1f408:
Merged as 3045c5e307dcedf116991c51c7cdbd81a2ffdfe5

PR #569, opened 2023-07-18
fix: bellhop emoji properly works as a ping react

rladenson:
Discord added a random variant selector

### DIFF
--- a/PluralKit.Bot/Handlers/ReactionAdded.cs
+++ b/PluralKit.Bot/Handlers/ReactionAdded.cs
@@ -86,7 +86,7 @@ public class ReactionAdded: IEventHandler<MessageReactionAddEvent>
         // Proxied messages only exist in guild text channels, so skip checking if we're elsewhere
         if (!DiscordUtils.IsValidGuildChannel(channel)) return;
 
-        switch (evt.Emoji.Name)
+        switch (evt.Emoji.Name.Split("\U0000fe0f", 2)[0])
         {
             // Message deletion
             case "\u274C": // Red X


PluralKit.Bot/Handlers/ReactionAdded.cs
```diff
@@ -1,3 +1,5 @@
+using System.Diagnostics.Tracing;
```
u1f408, line 1: This isn't needed, betting Visual Studio added it automatically for some reason?

PluralKit.Bot/Handlers/ReactionAdded.cs
```diff
@@ -108,7 +110,7 @@ private async ValueTask TryHandleProxyMessageReactions(MessageReactionAddEvent e
                 }
 
             case "\U0001F514": // Bell
-            case "\U0001F6CE": // Bellhop bell
+            case "\U0001F6CE\U0000fe0f": // Bellhop bell
```
u1f408, line 113: I'd rather that the variation selector gets stripped off entirely before we check these, instead of adding it to the checks whenever Discord break something!

I'd adjust the `evt.Emoji.Name` in the `switch` statement at the top of these checks to something like `evt.Emoji.Name.Split("\U0000fe0f", 2)[0]` so it always just gets the character(s) before the variation selector, if there is one.

u1f408 (changes requested):

PluralKit.Bot/Handlers/ReactionAdded.cs
rladenson: Hmm, do we need to worry about discord adding in other variation selectors? IE, should I just grab the first unicode id no matter what the second is? Or might that cause more issues down the line?

u1f408:
Merged in b75e3fe7d30c63ce25a0e34f7deabb47d72ecd55 😸

PR #572, opened 2023-07-18
docs: Add more info to member avatar set response when webhook avatar is set

rladenson:


u1f408:
Merged as d12cd481f6f94eebf22dfafb7c3720e0351f09ae

PR #575, opened 2023-07-26
Create LICENSE file

SomeAspy:
Add a license file to make it more clear to people what you can and cannot do with this repo. update accordingly with whatever names you want. Based off of the bottom header in readme.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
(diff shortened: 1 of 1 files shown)


u1f408:
See #545.

SomeAspy:
This doesn't make much sense to me- you don't want the license header purely because you don't want to distribute it?

u1f408:
We don't want to have text in the repo telling other people to use the AGPL (which parts of the license text effectively do), and we *especially* don't want to evangelize the Free Software Foundation in any way, given their (and Richard Stallman's) track record.

PR #578, opened 2023-08-02
feat: Remove double push notifications

rladenson:


rladenson:
I think the original conversation with Iris was to push this to the beta bot for a bit to see how it worked and then decide. I don't love it being a cfg setting bc it's for the convenience of the people getting the push notifs, not the people sending the messages.

u1f408 (approved):
lgtm! this is one of those changes where we won't know how annoyed people are with it until it's live, but i think it's worthwhile

PR #579, opened 2023-08-04
fix: add constraint for system name and avatar privacy

rladenson:


u1f408:
Merged as deeb3ad3a44b44465ae47453963255ce1944b5a7

PR #587, opened 2023-08-25
Add Regex message editing

nim-ka:
This PR adds the `-regex`/`-x` flag to `pk;edit`, which enables Regex message editing.

An example command looks like: `pk;edit -regex s/(\w+)(w)/$2$1/i`, which when applied to the message `I'm Drinking aterW` will edit the message to say `I'm Drinking Water`.

The expected syntax of the Regex input is `s|X|Y` or `s|X|Y|F`, where `|` is any character, `X` is a C# Regex string, `Y` is a substitution pattern, and `F` is an optional set of Regex flags (allowed flags are `gimnsx`). New possible error messages include:
"❌ You can't use the append or prepend options with a Regex."
"❌ Could not parse Regex. The expected formats are s|X|Y or s|X|Y|F, where | is any character, X is a valid Regex to search for matches of, Y is a substitution string, and F is a set of Regex flags."
"❌ Invalid Regex flag '&lt;character&gt;'. Valid flags include 'g', 'i', 'm', 'n', 's', and 'x'."
"❌ Regex took too long to run." (I haven't actually gotten this error to show in practice; it's there as a failsafe)


Lunaphied:
> I considered making the format s|X|Y, where | can be any single character and will be interpreted as a separator, like sed. Doing it this way allows for Regexes to contain forward slashes. (I think the current naive way I do it, a Regex with a forward slash would not be possible.) Didn't end up doing that, not sure if that's something that is wanted. 

Definitely would prefer if it had this personally. Additionally it would be good if either way the final trailing slash was optional, since it's not necessary when flags aren't being specified

nim-ka:
@Lunaphied There is no final separator used, it's `s/X/Y`. I'll look into documentation but afaik all the other options for pk;edit are equally undocumented? I might be wrong on that but I discovered them through the code and couldn't find any references to them elsewhere. So I don't know if a docs update would be in scope of this PR? If it is then I can do it.

u1f408:
Docs update would definitely be good in the PR! This is a very welcome change, I'll give this a proper review tomorrow.

nim-ka:
Noticed that -append/-prepend were documented (though not fully) in the Tips and Tricks page, so added documentation for the new Regex flags as well as existing pk;edit flags on there.

u1f408:
A quick note (since I haven't had time to fully review this yet): I would prefer if options like `-global` weren't flags in this case - it feels better to me to parse them off the end of the regex (if a final separator is present and there are characters after it) in the same way you'd specify them with `sed` etc

nim-ka:
That's true; but `g` would be the only one, as the other modifiers are done inline (e.g. `s/(?i)test/TEST`), so it didn't seem worth it to me.

nim-ka:
The more I think about it though the more I feel like just adding that sed separator thing mentioned earlier (`s|A|B`), and if I update the code for that I might just tack on recognizing `|g` on the end too...

nim-ka:
> The more I think about it though the more I feel like just adding that sed separator thing mentioned earlier (`s|A|B`, arbitrary `|`), and if I update the code for that I might just tack on recognizing `|g` on the end too...

Completed this change, and also added support for recognizing more flags that would otherwise be only accessible by setting them inline. I've edited the main post of this PR to reflect the new changes.

u1f408 (approved):

nim-ka:
lol, just did another look over this PR and noticed the tips-and-tricks.md page is broken because I didn't escape the vertical slash 😅 what should be done about that? is there an in-progress docs update that that can be tacked on to or nah

PR #589, opened 2023-09-11
table format fix

programing-monkey:
The row about regex in misc flags table of `docs/content/tips-and-tricks.md` was missing part of it's last column due to some unescaped pipes.

(this is my first pull request, let me know if I've done something wrong)


u1f408:
Thanks! :)

PR #599, opened 2024-01-18
feat(api): add endpoint for latch clear

ashen-dawn:
## What changes are in this PR?
- Adds a `/systems/@me/autoproxy/unlatch`endpoint, as an API equivalent to the existing `\\` message command
- Adds documentation for the endpoint

## Why is this useful?
My system also uses and contributes to [Séance](https://github.com/Qyriad/Seance), and though we like to keep things *mostly* consistent, situations arise where not all of our bot instances are in every server, so we end up using both PluralKit and Séance in the same channels, and managing autoproxy between both of those can get kind of messy.

When we switch from using a Séance instance to Pluralkit that goes fine - our fork's autoproxy implementation in Séance is built with the assumption that there will be other bot instances, so we just need to make sure the "peer pattern" regex each bot has also covers our Pluralkit message patterns.  But at the moment, there is no way for Pluralkit to know that it should unlatch when a Séance instance starts proxying.  Adding this endpoint will allow us to build that functionality into Séance, giving us a bit of an easier time in servers where we're using both.

This is admittedly a bit of an out-there edge use case, but as it's functionality the bot already had, I figured adding an API endpoint to expose that functionality programmatically would not be an issue.

### Potential other options
We could *theoretically* get the behavior we need by querying the current autoproxy state, turning autoproxy off, and then putting it back . . . but that would mean making three API requests in quick succession, which would exceed your rate limit

For a while I thought maybe I'd be able to send a PATCH request to the existing `/systems/@me/autoproxy` endpoint to just reset the latch state, but it doesn't appear to work that way and it seemed like better API design to give this its own endpoint if we were going to add the functionality in

## Potential issues
With the way I've got my code right now, the API routing itself *does* allow for GET and PATCH requests to the `unlatch` endpoint, and POST requests to the base `autoproxy` endpoint.  As it looks like you're running the API behind a Rust proxy I figured that would be fine, and I preferred to follow the code pattern you've already got in place for routing (I'm not *super* familiar with .NET so extending it was beyond the scope of what I could do myself easily), but if that's a problem let me know

### DIFF
--- a/PluralKit.API/Controllers/v2/AutoproxyControllerV2.cs
+++ b/PluralKit.API/Controllers/v2/AutoproxyControllerV2.cs
@@ -22,6 +22,10 @@ public class AutoproxyControllerV2: PKControllerBase
     public Task<IActionResult> PatchWrapper([FromRoute] string systemRef, [FromQuery] ulong? guild_id, [FromQuery] ulong? channel_id, [FromBody] JObject? data)
         => Entrypoint(systemRef, guild_id, channel_id, data);
 
+    [HttpPost("systems/{systemRef}/autoproxy/unlatch")]
+    public Task<IActionResult> PostWrapper([FromRoute] string systemRef, [FromQuery] ulong? guild_id, [FromQuery] ulong? channel_id)
+        => Entrypoint(systemRef, guild_id, channel_id, null);
+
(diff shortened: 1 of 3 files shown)


ashen-dawn:
Closing in favor of #615

PR #611, opened 2024-02-07
Fix: avoid 403 status codes on image upload

Draconizations:
Avatar/banner uploads to discord now fetch with their query params included, which avoids those 403 status codes that we've been plagued with the past day.

(Even though they just seemed to have fixed themselves, can't hurt to be prepared for an inevitable future breakage)

Attachment URLs in the db are stored without these query params.

### DIFF
--- a/PluralKit.Bot/CommandSystem/Context/ContextAvatarExt.cs
+++ b/PluralKit.Bot/CommandSystem/Context/ContextAvatarExt.cs
@@ -52,7 +52,7 @@ public static class ContextAvatarExt
 public struct ParsedImage
 {
     public string Url;
-    public string CleanUrl;
+    public string? CleanUrl;
     public AvatarSource Source;
     public User? SourceUser;
 }
--- a/PluralKit.Bot/CommandSystem/Context/ContextAvatarExt.cs
+++ b/PluralKit.Bot/CommandSystem/Context/ContextAvatarExt.cs
@@ -33,11 +33,14 @@ public static class ContextAvatarExt
         // If we have an attachment, use that
         if (ctx.Message.Attachments.FirstOrDefault() is { } attachment)
         {
-            // XXX: strip query params from attachment URLs because of new Discord CDN shenanigans
+            // XXX: discord attachment URLs are unable to be validated without their query params
+            // keep both the URL with query (for validation) and the clean URL (for storage) around
             var uriBuilder = new UriBuilder(attachment.ProxyUrl);
-            uriBuilder.Query = "";
 
-            return new ParsedImage { Url = uriBuilder.Uri.AbsoluteUri, Source = AvatarSource.Attachment };
+            ParsedImage img = new ParsedImage { Url = uriBuilder.Uri.AbsoluteUri, Source = AvatarSource.Attachment };
+            uriBuilder.Query = "";
+            img.CleanUrl = uriBuilder.Uri.AbsoluteUri;
+            return img;
         }
 
         // We should only get here if there are no arguments (which would get parsed as URL + throw if error)
@@ -49,6 +52,7 @@ public static class ContextAvatarExt
 public struct ParsedImage
 {
     public string Url;
+    public string CleanUrl;
     public AvatarSource Source;
     public User? SourceUser;
 }
--- a/PluralKit.Bot/Commands/Groups.cs
+++ b/PluralKit.Bot/Commands/Groups.cs
@@ -263,7 +263,7 @@ public class Groups
 
             await AvatarUtils.VerifyAvatarOrThrow(_client, img.Url);
 
-            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { Icon = img.Url });
+            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { Icon = img.CleanUrl });
 
             var msg = img.Source switch
             {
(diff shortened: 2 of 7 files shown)


PluralKit.Bot/Commands/Groups.cs
```diff
@@ -263,7 +263,7 @@ async Task SetIcon(ParsedImage img)
 
             await AvatarUtils.VerifyAvatarOrThrow(_client, img.Url);
 
-            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { Icon = img.Url });
+            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { Icon = img.CleanUrl });
```
u1f408, line 266: this will break setting avatars using a URL as text in a bot command - `img.CleanUrl || img.Url`

PluralKit.Bot/CommandSystem/Context/ContextAvatarExt.cs
```diff
@@ -49,6 +52,7 @@ public static class ContextAvatarExt
 public struct ParsedImage
 {
     public string Url;
+    public string CleanUrl;
```
u1f408, line 55: needs to be nullable - `public string? CleanUrl;`

PluralKit.Bot/Commands/Groups.cs
```diff
@@ -328,7 +328,7 @@ async Task SetBannerImage(ParsedImage img)
 
             await AvatarUtils.VerifyAvatarOrThrow(_client, img.Url, true);
 
-            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { BannerImage = img.Url });
+            await ctx.Repository.UpdateGroup(target.Id, new GroupPatch { BannerImage = img.CleanUrl });
```
u1f408, line 331: same deal here

PluralKit.Bot/Commands/MemberAvatar.cs
```diff
@@ -139,7 +140,7 @@ public async Task WebhookAvatar(Context ctx, PKMember target)
 
         ctx.CheckSystem().CheckOwnMember(target);
         await AvatarUtils.VerifyAvatarOrThrow(_client, avatarArg.Value.Url);
-        await UpdateAvatar(location, ctx, target, avatarArg.Value.Url);
+        await UpdateAvatar(location, ctx, target, avatarArg.Value.CleanUrl);
```
u1f408, line 143: and here

PluralKit.Bot/Commands/MemberEdit.cs
```diff
@@ -182,7 +182,7 @@ async Task SetBannerImage(ParsedImage img)
         {
             await AvatarUtils.VerifyAvatarOrThrow(_client, img.Url, true);
 
-            await ctx.Repository.UpdateMember(target.Id, new MemberPatch { BannerImage = img.Url });
+            await ctx.Repository.UpdateMember(target.Id, new MemberPatch { BannerImage = img.CleanUrl });
```
u1f408, line 185: and here

PluralKit.Bot/Commands/SystemEdit.cs
```diff
@@ -475,7 +475,7 @@ async Task SetIcon(ParsedImage img)
 
             await AvatarUtils.VerifyAvatarOrThrow(_client, img.Url);
 
-            await ctx.Repository.UpdateSystem(target.Id, new SystemPatch { AvatarUrl = img.Url });
+            await ctx.Repository.UpdateSystem(target.Id, new SystemPatch { AvatarUrl = img.CleanUrl });
```
u1f408, line 478: and all 3 changes in this file

PluralKit.Bot/Utils/AvatarUtils.cs
```diff
@@ -63,11 +64,24 @@ public static async Task VerifyAvatarOrThrow(HttpClient client, string url, bool
     // This lets us add resizing parameters to "borrow" their media proxy server to downsize the image
     // which in turn makes it more likely to be underneath the size limit!
     private static readonly Regex DiscordCdnUrl =
-        new(@"^https?://(?:cdn\.discordapp\.com|media\.discordapp\.net)/attachments/(\d{17,19})/(\d{17,19})/([^/\\&\?]+)\.(png|jpg|jpeg|webp)(\?.*)?$");
+        new(@"^https?://(?:cdn\.discordapp\.com|media\.discordapp\.net)/attachments/(\d{17,19})/(\d{17,19})/([^/\\&\?]+)\.(png|jpg|jpeg|webp)(?<query>\?.*)?$");
```
u1f408, line 67: i'd suggest changing the query capture here to not capture the question mark (something like `(?:\?(?<query>.*))?`) - makes rewriting the query string easier. 

with that, the `DiscordMediaUrlReplacement` string can stay the same as it was before this - and then change `TryRewriteCdnUrl` to something like this (untested, i'm writing this in the review comment field)

```csharp
public static string? TryRewriteCdnUrl(string? url) {
    if (url == null)
        return null;

    var match = DiscordCdnUrl.Match(url);
    var newUrl = DiscordCdnUrl.Replace(url, DiscordMediaUrlReplacement);
    if (match.Groups["query"].Success)
        newUrl += "&" + match.Groups["query"].Value;

    return newUrl;
}
```

PluralKit.Bot/Utils/AvatarUtils.cs
```diff
@@ -26,6 +26,7 @@ public static async Task VerifyAvatarOrThrow(HttpClient client, string url, bool
             throw new PKError("Due to server issues, PluralKit is unable to read images hosted on toyhou.se.");
 
         url = TryRewriteCdnUrl(url);
+        Console.WriteLine(url);
```
u1f408, line 29: remove this please!

u1f408 (changes requested):

u1f408 (approved):

PR #615, opened 2024-02-10
fix(api): refactor autoproxy controller so unlatching can work

ashen-dawn:
Follow-up to #599 after it was pointed out in the discord that this was supposed to already be possible with the current patch endpoint.
- Fixes the 500 error that was being caused then sending a patch of `{"autoproxy_member": null}` in latch mode - that now properly unlatches
- Clarifies the error handling around invalid states (setting it to mode `member` with no member specified and setting a member when the system's settings for that server is not in mode `member`)

I admittedly didn't *entirely* follow the way the logic was working before so while I believe this works the same (aside from the additional error case), and I have tested it and it appears working, I wouldn't be surprised if I missed something


ashen-dawn:
Ah yep knew I forgot something - that has been run

PR #623, opened 2024-03-07
Add shorthand 'pk;x' for 'pk;edit -regex'

nim-ka:
Also adds entries for `pk;e` and `pk;x` in the command shorthands section of the Tips and Tricks.


AtelierSnek:
The command list should potentially also be updated, as it currently doesn't mention that `-regex` is even an option.

nim-ka:
Potentially agreed, but it also hadn't mentioned other pk;edit flags (e.g. -append, -prepend); those are only in the Tips & Tricks section of the docs so thats the only place I added documentation for regex editing stuff

AtelierSnek:
Hmm. Fair point. Might be worth adding the flags into the command list in the general case. We'll look into creating a PR for that.

PR #631, opened 2024-05-02
feat(api): always set system id on members/groups

z0w13:
This pull request sets the `system` property on members/groups across all API responses.

Reason for this change is so that all API responses are consistent in having the `system` field set.

In some cases we don't need it (i.e. `/v2/systems/{systemRef/switches/{switchRef}`) but it saves an extra API call in other cases, you'd need to separately query the group after `GET /groups/{groupRef}/members` to get the system ID.

So we figured to just make it consistent across the board


u1f408:
thanks! 😸

PR #655, opened 2024-08-15
Escape Discord usernames in system embed.

### LINKED ISSUE #641: Escaping formatting syntax from usernames
When I request info about a alter by replying with ❓ to a message it applies formatting syntax. In this case it makes part of the name italic as can be seen in the screenshot below.
image

Looking at the users profile it shows 2 "\_" around the italic parts.
image
this is likely due to pluralkit not escaping formatting like "\_" in usernames.

threeoh6000:
This fixes #641 by escaping usernames correctly while preserving the inline link to the user's profile.

Please bear with me if I did anything wrong in terms of code or syntax standards. I'm more accustomed to working with Rust and most of my C# experience is with Unity.

Thank you!

### DIFF
--- a/PluralKit.Bot/Services/EmbedService.cs
+++ b/PluralKit.Bot/Services/EmbedService.cs
@@ -43,7 +43,7 @@ public class EmbedService
     {
         // Fetch/render info for all accounts simultaneously
         var accounts = await _repo.GetSystemAccounts(system.Id);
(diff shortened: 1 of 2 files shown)


threeoh6000:
Ah, apologies! I've just done this and tested it works. I was a bit hesitant to modify it in case I broke other stuff 😅

PR #665, opened 2024-09-03
feat: add more detailed switch editing and switch copying

rladenson:
[Notion Link](https://www.notion.so/pk-switch-flags-to-allow-for-partial-editing-ecd50b2e33fc4b20b18a341fe091d580)

Note: I should probably add documentation to this PR...


rladenson:
OK there's the docs. There isn't anything about switch editing in the user guide btw, but I can't formulate that whole thing right now.

PR #668, opened 2024-09-13
Rust command parser

avioletheart:
this is a rewrite of the command parser in rust, defining all commands statically and creating at build time both a command tree and documentation structure that can be used both in the bot and to generate the command list on the online docs.

commands are defined with a custom-ish DSL parsed by the Rust compiler (yay proc macros!) that looks like this:
```rust
commands! {
    category(system, "System commands", "To target a specific system, replace [...]"):

    command(+system, system_view_self, "Shows information about your system");
    command(+system @SystemRef, system_view, "Shows information about a system");
    command(+system @SystemRef description @FullString, member_desc_update, "Changes your system's description");
    command(+system @SystemRef privacy, system_privacy_view, "Displays your system's current privacy settings");
    command(+system @SystemRef privacy @SystemPrivacyTarget @PrivacyLevel, "Changes your system's privacy settings");
```

an explanation of the tokens in `command` definitions:
- three comma-separated sections: a list of tokens to match a command string with, a (string) function callback, and a description
- a plain `text` is a bot-defined part of a command, say `member` in `pk;member alyssa`.
- `+text` is a collection of aliases; here `"system", "s"`
- `@text` is a collection of values that are used as _command arguments_
  - this can either match anything (in the case of a description string) or only specific values (privacy targets/levels)

the code is somewhat complex so i intend to document this as much as possible. (currently only some code is documented)

things left to do:
- the `todo!` items listed in code
- add all commands
- command aliases
- command categories
- multi-value token matching (used in `pk;cfg`)
- generate docs
- update .net docker builds
- update c# code to take static arguments rather than using `ctx.Match`
- more?

### DIFF
--- a/lib/command_system_macros/src/lib.rs
+++ b/lib/command_system_macros/src/lib.rs
@@ -1,57 +1,95 @@
-use proc_macro2::{Delimiter, TokenStream, TokenTree, Ident, Literal, Span};
-use quote::quote;
+use proc_macro2::{Delimiter, TokenStream, TokenTree, Literal, Span};
+use syn::parse::{Parse, ParseStream, Result as ParseResult};
+use syn::{parse_macro_input, Token, Ident};
+use quote::{quote, quote_spanned};
 
-fn make_command(
-    tokens: Vec<TokenStream>,
-    help: Literal,
-    cb: Literal,
-) -> TokenStream {
-    quote! {
-        Command { tokens: vec![#(#tokens),*], help: #help.to_string(), cb: #cb.to_string() }
+enum CommandToken {
+    /// "typed argument" being a member of the `Token` enum in the
+    /// command parser crate.
+    ///
+    /// prefixed with `@` in the command macro.
+    TypedArgument(Ident, Span),
+
+    /// interpreted as a literal string in the command input.
+    ///
+    /// no prefix in the command macro.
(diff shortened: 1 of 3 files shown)


lib/command_system_macros/src/lib.rs
u1f408, line 1: the use of string `format!` for TokenStream generation *horrifies* me, please use [`syn`](https://lib.rs/crates/syn)'s `quote!` or something

lib/commands/src/token.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub enum TokenMatchResult {
+    NoMatch,
+    /// Token matched, optionally with a value.
+    Match(Option<String>),
+}
+
+// move this somewhere else
+lazy_static::lazy_static!(
+    static ref MEMBER_PRIVACY_TARGETS: Vec<String> = vec![
```
u1f408, line 34: spitballing: should privacy targets etc be macro-generated enums, with a token type for referencing them in the command structure DSL? even if they just get pushed through FFI as a string argument in the end, it'd make the whole "self-documenting" goal easier

u1f408 (commented):
i haven't looked *too* closely at this yet, but here's some initial thoughts

PR #674, opened 2024-10-03
feat: give better error message when people try to update their own member limits

rladenson:
I was unsure about which mockup I like better, I implemented the first but can switch it to the second
image


rladenson:
Just force-pushed, it uses mockup 2 now (plus a mention of the #bot-support channel)

PR #675, opened 2024-10-03
fix: description and name privacy respected while searching

rladenson:


rladenson:
Hmmm, is this an issue for groups as well?

rladenson:
It was an issue for groups as well. Added!

PR #681, opened 2024-10-07
fix: name privacy while searching inconsistent with other name privacy

rladenson:
Note: untested, will test when I can


rladenson:
Tested, seems to work properly!

PR #684, opened 2024-10-15
Add License to Codebase

tyrantlink:
Include the License along with the code, instead of just at the bottom of the README


tyrantlink:
found #545

PR #685, opened 2024-10-19
Added Maki logging cleanup support (maki.gg)

ijsbol:
Heya!

I'm one of the developers for https://maki.gg.

We're currently rolling out a change to our deleted message logger that includes "Message ID: [id]" in the embed footer. We added this so PluralKit (and other similar bots) will be able to scan these logs and delete them where necessary to stop log spam :)

I believe I've managed to implement the regex & log scanning correctly - though this is my first time contributing to PK so any advice is appreciated! 

img


u1f408 (approved):

u1f408:
thanks!! :)

PR #695, opened 2024-11-08
feat: cfg option to log a switch when proxying

rladenson:
Note: This ignores message deletions (including by reproxy). Adding in reproxy support may be possible but seems like way more trouble than it's worth (you have to figure out if the switch was triggered by the message being reproxied or by something else), but I don't think deleted message support is feasible at all.


rladenson:
Changing to a draft rq bc I forgot to add docs to this PR

PR #705, opened 2024-11-19
Add raw and plaintext formats to images

rladenson:
Will mark PR as ready to review once I have all image commands in it.


rladenson:
Ok the only thing missing now should be member banners

PR #747, opened 2025-06-28
Mark member guild settings model's `guild_id` as optional

00-kat:
I'm not certain if this is correct (please close this PR if it isn't), but every other instance where `guild_id`'s notes contains “only sent if the guild ID isn't already known” is optional.  (Plus, “only sent” sort of implies that it's optional.)


00-kat:
Yeah, in my opinion documenting it on the dispatch page (and perhaps cross-referencing it from here) would be a better idea.

PR #777, opened 2026-01-14
feat: add more options for when to make proxied messages @silent

### LINKED ISSUE #770: Add "match trigger message" mode to `pk;scfg suppress notifications`
`pk;scfg suppress notifications` is currently boolean - either "proxied messages are never `@silent`" (`off`), or "proxied messages are always `@silent`" (`on`). Add a third state to make proxied messages use the same value for `SUPPRESS_NOTIFICATIONS` as the trigger message. 

- Maybe rename the existing states also to make the behaviour clearer? (`on` -> `always`; `off` -> `never`)
- Should this also have a "invert state of trigger message" mode? (trigger message has `SUPPRESS_NOTIFICATIONS` -> proxied message does not; normal trigger message -> proxied message has `SUPPRESS_NOTIFICATIONS` flag)

This is a followup to https://github.com/PluralKit/PluralKit/issues/596#issuecomment-3283421250 after discussions in the support server - see https://canary.discord.com/channels/466707357099884544/1319461648251097288/1415892281948508180 for that discussion

rladenson:
This may conflict with #776


rladenson:
....I did not mean to ping someone I am so sorry

PR #791, opened 2026-07-15
fix: msg author can't always cancel yesno prompts

rladenson:
https://app.notion.com/p/Can-t-cancel-pk-link-mention-as-the-command-runner-314504c469a18021bd85e81ec8ae0781?v=eba10f3739224ef8824e5de90722a60c&source=copy_link

I don't think there are currently YesNoPrompts this affects other than linking but I figured it could possibly affect others in the future and there's no situation it would be bad in.


rladenson:
Wait this only accounts for buttons

